PR #19, opened 2025-05-01
cmd: update after library update

runcom:
@ben-krieger ptal


ben-krieger (approved):
LGTM. Thanks for updating this. We still need to do a whole refactor on go-fdo-server at some point to add auth, additional PRI APIs, and change to cobra for flag parsing. Let me know if you're going to do anything like that and we can work together, since I'll probably need to do it for another project soon.

cmd/fdo_server/server.go
```diff
@@ -628,7 +671,7 @@ func ownerModules(ctx context.Context, guid protocol.GUID, info string, chain []
 		if cmdDate && slices.Contains(modules, "fdo.command") {
 			if !yield("fdo.command", &fsim.RunCommand{
 				Command: "date",
-				Args:    []string{"--utc"},
+				Args:    []string{"+%s"},
```
ben-krieger, line 674: Any particular reason for this change?

runcom:
> and change to cobra for flag parsing

I'm on this full time now!

cmd/fdo_server/server.go
runcom: I don't know how that went in actually :confused:

runcom: https://github.com/fido-device-onboard/go-fdo-server/pull/20

PR #21, opened 2025-05-13
Move to Cobra

### LINKED ISSUE #9: Use a better flag parsing library than stdlib
This is the very same issue tracking as https://github.com/fido-device-onboard/go-fdo-client/issues/6

> This code is very messy, because it started from the go-fdo library example. The example was meant to be simple (though it probably gained too many features to still be called that), so it did not use Cobra or anything.

> @ben-krieger are you open to also use cobra in go-fdo-server? it's pretty confusing over there, especially since a single binary is handling different server roles and doesn't allow for proper configuration files to be setup per-role

runcom:
Closes #9


runcom:
@ben-krieger opened this draft to gauge the direction of the cli UI with cobra. I didn't want to go all in w/o feedback - the PR aims to translate 1-1 with what we have today without disrupting too much the experience and then follow up to include the move to a proper config file, remove db flags etc etc

cmd/root.go
```diff
@@ -0,0 +1,71 @@
...
+// 	[--external-addr=<address>]
+// 	[--command-date]
+// 	[--command-wget=<url>]
+// 	[--command-upload=<file>]
+// 	[--command-upload-dir=<path>]
+// 	[--command-download=<file>]
+// 	[--reuse-credentials]
+// 	--db <path>
+// 	--db-pass <password>
+// 	<http-addr>
```
runcom, line 35: @ben-krieger a draft of what this looks like

cmd/fdo_server/server.go
```diff
@@ -161,6 +161,7 @@ func (s *Server) Start() error {
 		return err
 	}
 	defer func() { _ = lis.Close() }()
+	// TODO(runcom): we don't listen on external address?
```
ben-krieger, line 164: The external address represents "how will a device connect to this service?" It could be an FQDN or IP (either public or private, depending on the deployment scenario). This is then used in the rendezvous info for the device credential (I'm guessing).

cmd/importVoucher.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+// importVoucherCmd represents the importVoucher command
+var importVoucherCmd = &cobra.Command{
+	Use:   "importVoucher",
```
ben-krieger, line 15: commands should be kebab-case, right?

cmd/importVoucher.go
```diff
@@ -0,0 +1,40 @@
+/*
+Copyright © 2025 NAME HERE <EMAIL ADDRESS>
+
+*/
```
ben-krieger, line 4: Let's use the same SPDX header style as in the existing files

cmd/importVoucher.go
```diff
@@ -0,0 +1,40 @@
...
+package cmd
```
ben-krieger, line 5: Make sure to have an empty line above package or the copyright stuff will end up as package docs!

main.go
```diff
@@ -0,0 +1,11 @@
+/*
+Copyright © 2025 NAME HERE <EMAIL ADDRESS>
+
+*/
+package main
```
ben-krieger, line 5: I think we should actually move main.go into the project root so that the command can be installed with `go install github.com/fido-device-onboard/go-fdo-server@latest`.

cmd/root.go
ben-krieger: Looks good. I would skip the `--command-XXX` flags. Even though that's existing functionality, we're going to want something more powerful.

I'm not exactly sure what that'll look like, so when we get there, please rely on me to help define and implement it. I think we'll likely want a whole subcommand for getting/setting complex per-device FSIM chains...

runcom: agreed, I've initially added those --command-XXX just for the sake of supportability with the _old_ version - I think we might just keep them in this PR iteration so there's feature parity and then move away asap with a follow up (just for the sake of moving fast and share work)

cmd/importVoucher.go
runcom: yeah, figured that afterwards :smile:

main.go
ben-krieger: ... Yes. GitHub's UI confused me.

runcom:
pushed the 3 subcommands for resell/import-voucher/print-owner-pubkey - please let me know how do they look while I finish up the serve subcommand :)

README.md
ben-krieger, line 1: I suggest removing the Makefile and update-deps.sh script and then updating the corresponding README sections.

cmd/printOwnerPubkey.go
```diff
@@ -0,0 +1,50 @@
...
+			Type:  "PUBLIC KEY",
+			Bytes: der,
+		})
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(printOwnerPubkeyCmd)
+	printOwnerPubkeyCmd.Flags().StringVar(&pubKeyTypeS, "type", "", "Public key type")
+	printOwnerPubkeyCmd.MarkFlagRequired("type")
```
ben-krieger, line 49: There needs to be a key-size flag for RSA keys (required for PKCS/PSS, ignored otherwise).

cmd/importVoucher.go
```diff
@@ -0,0 +1,74 @@
...
+		if err := cbor.Unmarshal(blk.Bytes, &ov); err != nil {
+			return fmt.Errorf("error parsing voucher: %w", err)
+		}
+
+		// Check that voucher owner key matches
+		expectedPubKey, err := ov.OwnerPublicKey()
+		if err != nil {
+			return fmt.Errorf("error parsing owner public key from voucher: %w", err)
+		}
+		ownerKey, _, err := state.OwnerKey(context.Background(), ov.Header.Val.ManufacturerKey.Type, 3072)
```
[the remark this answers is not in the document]
runcom: will file an issue

Makefile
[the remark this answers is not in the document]
runcom: I'll remove the makefile indeed in this PR, will update asap

cmd/importVoucher.go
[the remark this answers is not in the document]
runcom: this is also related to https://github.com/fido-device-onboard/go-fdo-server/issues/22 so I think we just need an issue for voucher verification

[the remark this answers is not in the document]
runcom: https://github.com/fido-device-onboard/go-fdo-server/issues/24

runcom:
all comments addressed

runcom:
> Approving this for merge, but going forward files should be named with snake_case instead of camelCase.

ah, gotcha, I overlooked that, it's autogenerated like that by the cobra-cli command

PR #53, opened 2025-08-26
feat: enhance the vouchers api

runcom:
Basically have it list all vouchers with their fields too so an admin can have a better look at what it has.
This is also needed to better export vouchers, for example, somebody provisions a system and knows the MAC used as device_info and they can thus just grab that directly (real world use case). Just relying on GUID is not very handy as when a device is initialized and a voucher created in the db, there's no way to link them together.

~~This patch needs https://github.com/fido-device-onboard/go-fdo/pull/139~~


api/handlers/vouchers.go
```diff
@@ -29,26 +31,63 @@ import (
...
+		}
+		w.Header().Set("Content-Type", "application/json")
+		if err := json.NewEncoder(w).Encode(out); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
 		return
 	}
 
-	voucher, err := db.FetchVoucher(guid)
+	// Filters provided: return a single voucher as PEM
```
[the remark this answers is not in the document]
runcom: so, what if we had the single voucher return a JSON struct with "guid": "PEM_VOUCHER"? this way we stay consistent on output (although, the upload code in the owner server needs a PEM_VOUCHER directly, which may be fine I guess)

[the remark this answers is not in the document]
runcom: figured it wasn't that pretty... now it's more REST

`/vouchers + query param` -> always give you a json array (what if somebody messes up with device_info and it's duplicate)
`/vouchers/GUID` -> gives you the pem voucher as it's should be always just one

internal/to0/to0.go
```diff
@@ -18,6 +18,7 @@
 )
 
 func RegisterRvBlob(RvInfo [][]protocol.RvInstruction, to0Guid string, voucherState fdo.OwnerVoucherPersistentState, keyState fdo.OwnerKeyPersistentState, useTLS bool) error {
+	// TODO: use protocol.ParseOwnerRvInfo()
```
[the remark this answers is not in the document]
runcom: I'll create an issue for this

api/handlers/vouchers.go
```diff
@@ -29,40 +30,104 @@ import (
...
+	}
+	out := make([]voucherView, 0, len(vouchers))
+	for _, v := range vouchers {
+		out = append(out, voucherView{
+			GUID:       fmt.Sprintf("%x", v.GUID),
+			DeviceInfo: v.DeviceInfo,
+			CreatedAt:  v.CreatedAt.UnixMicro(),
+			UpdatedAt:  v.UpdatedAt.UnixMicro(),
+		})
+	}
```
mmartinv, line 105: Why do we need this? Why don't we just return the serialized vouchers returned by `db.queryVouchers()`?

mmartinv (commented):
Lots of duplicated and unneeded code, please take a look at the comments.

api/handlers/vouchers.go
runcom: even if we omitempty on the cbor field from db.Voucher, we still have guid in []byte instead of hex - this is just a supporting struct for the view at the api level

api/handlers/vouchers.go
```diff
@@ -29,40 +30,104 @@ import (
...
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(out); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
```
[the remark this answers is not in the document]
runcom: thanks, removed all the duplications

api/handlers/vouchers.go
mmartinv: Honestly, I think it would be better to redefine the GUID field in the Voucher struct so we don't need to walk the vouchers again, something like:
```diff
diff --git a/internal/db/models.go b/internal/db/models.go
index a1e0936..7a580aa 100644
--- a/internal/db/models.go
+++ b/internal/db/models.go
@@ -3,15 +3,34 @@
 
 package db
 
-import "time"
+import (
+       "encoding/hex"
+       "encoding/json"
+       "time"
+)
 
 type Data struct {
        Value interface{} `json:"value"`
 }
 
+type GUID []byte
+
+func (t *GUID) UnmarshalJSON(b []byte) (err error) {
+       var g string
+       if err = json.Unmarshal(b, &g); err != nil {
+               return
+       }
+       *t, err = hex.DecodeString(g)
+       return
+}
+
+func (t *GUID) MarshalJSON() (b []byte, err error) {
+       return json.Marshal(hex.EncodeToString(*t))
+}
+
 type Voucher struct {
-       GUID       []byte    `json:"guid"`
-       CBOR       []byte    `json:"cbor"`
+       GUID       GUID      `json:"guid"`
+       CBOR       []byte    `json:"cbor,omitempty"`
        DeviceInfo string    `json:"device_info"`
        CreatedAt  time.Time `json:"created_at"`
        UpdatedAt  time.Time `json:"updated_at"`
```

internal/db/db.go
```diff
@@ -51,10 +52,108 @@ func createOwnerInfoTable() error {
...
+	// If there is a second row, error out due to ambiguity
+	if rows.Next() {
+		return Voucher{}, fmt.Errorf("multiple vouchers matched filters")
+	}
+
+	return voucher, nil
+}
+
+// QueryVouchers returns all owner vouchers matching optional filters (without CBOR), ordered by updated_at DESC.
+func QueryVouchers(filters map[string]interface{}) ([]Voucher, error) {
```
mmartinv, line 118: Same here, lot of duplicated code: `FetchVoucher` is pretty much the same function as `QueryVouchers`. I suggest to redefine the `QueryVouchers` function so it takes an additional boolean argument to determine if the CBOR field must be included in the resulting vouchers or not and call it from `FetchVoucher` and there check if the result's size is greater than 1 to detect duplicates.

mmartinv (commented):
Just a couple of minor changes, please take a look

internal/db/db.go
runcom: right, I think we don't even need the cbor boolean as omitempty seems enough (and we don't populate that in QueryVouchers)

runcom:
> I think we shouldn't retrieve the CBOR values from database if not needed.

ftr, I don't like the double scan (and it grows quickly to me if we add stuff). I'd rather retrieve everything as we don't have tera of data and the filter out in code.

But I've changed it, no biggie

Issue #55, opened 2025-08-27
RegisterRvBlob should use the go-fdo.protocol pkg


runcom:
Right now most of the things dealing with RvInfo have been "reimplemented" in the rvinfo package but it should be possible to use the `protocol` subpkg from go-fdo. This issue should assess the current code and make changes accordingly to just use the go-fdo protocol package.

Issue #38, opened 2025-07-18
Fix rvinfo hanlding during voucher import


runcom:
Right now the rvinfo for vouchers are part of the http handlers. Following https://github.com/fido-device-onboard/go-fdo-server/pull/28#discussion_r2210415313 we need to make sure to refactor it so that when the ov is imported to0 is run too and get rid of the global rvinfo in the http handler.

Issue #29, opened 2025-07-01
Panic when POSTing rvinfo directive that does not contain an RVIPAddress entry


kgiusti:
### Nature of the bug

Panic occurs if RVIPAddress is not present when POSTing RV Info directive.

### Steps taken to find bug

Run the fdo server:
`./fdo_server -debug -http 127.0.0.1:8038 -db ./mfg.db -db-pass "TrustNo1!"`

Post the following RV directive:
`curl -D - --location --request POST 'http://localhost:8038/api/v1/rvinfo' --header 'Content-Type: text/plain' --data-raw '[[[0], [12,1],[5,"localhost"],[3,8043]]]'`

Note the RVIPAddress is not given but the RVDns address is.

### Steps to ensure bug is gone

The above curl POST command should succeed.

### Optional: Reproduction

See above

### Client Debug Logs

None.

### Server Debug Logs

```console
[14:39:27] DEBUG: Received RV request
  method: POST
  path: /api/v1/rvinfo
[14:39:27] DEBUG: rvData created
[14:39:27] DEBUG: RV ->
  index: 0
  key: RVDevOnly
  value: <nil>
[14:39:27] DEBUG: RV ->
  index: 0
  key: RVProtocol
  value: 1
[14:39:27] DEBUG: RV ->
  index: 0
  key: RVDns
  value: localhost
[14:39:27] DEBUG: RV ->
  index: 0
  key: RVDevPort
  value: 8043
[14:39:27] INFO: http: panic serving 127.0.0.1:51444: interface conversion: interface {} is nil, not string
goroutine 35 [running]:
net/http.(*conn).serve.func1()
	/usr/lib/golang/src/net/http/server.go:1947 +0xbe
panic({0x925dc0?, 0xc000afe4b0?})
	/usr/lib/golang/src/runtime/panic.go:791 +0x132
github.com/fido-device-onboard/go-fdo-server/internal/rvinfo.UpdateRvInfo(0xc00207ec40, 0x0, 0xc000afe420)
	/home/kgiusti/tmp/go-fdo-server/internal/rvinfo/rvinfo.go:119 +0x1194
github.com/fido-device-onboard/go-fdo-server/internal/rvinfo.RetrieveRvInfo(0xc00207ec40)
	/home/kgiusti/tmp/go-fdo-server/internal/rvinfo/rvinfo.go:57 +0x23f
[...]

Issue #14, opened 2025-02-13
Handle JSON Explicitly


ben-krieger:
Currently the Java-compatible API parses JSON into an `interface{}` and then marshals it again before storing it in the DB (why even parse, then?). The structure parsed should be explicit. Also, there should be no panicing possible as there currently is with the MustMarshal-type utility functions.

PR #57, opened 2025-08-29
refactor rvinfo and ownerinfo

runcom:
saving some LOC plus a better handling of the data 
the patch cut lines of code in half basically (the additions are test cases)

closes https://github.com/fido-device-onboard/go-fdo-server/issues/29
closes https://github.com/fido-device-onboard/go-fdo-server/issues/55
closes https://github.com/fido-device-onboard/go-fdo-server/issues/38
closes https://github.com/fido-device-onboard/go-fdo-server/issues/14


internal/db/db.go
```diff
@@ -141,3 +144,123 @@ func FetchData(tableName string) (Data, error) {
...
+	var raw any
+	if err := json.Unmarshal([]byte(value), &raw); err != nil {
+		return nil, fmt.Errorf("error unmarshalling owner_info: %w", err)
+	}
+	items, ok := raw.([]any)
+	if !ok {
+		return nil, fmt.Errorf("invalid owner_info format: outer not array")
+	}
+
+	out := make([]protocol.RvTO2Addr, 0, len(items))
```
runcom, line 216: the parsing below may belong to the protocol package in go-fdo

internal/db/db.go
```diff
@@ -141,3 +144,123 @@ func FetchData(tableName string) (Data, error) {
...
+	if err := json.Unmarshal([]byte(value), &raw); err != nil {
+		return nil, fmt.Errorf("error unmarshalling rvInfo: %w", err)
+	}
+
+	outer, ok := raw.([]any)
+	if !ok {
+		return nil, fmt.Errorf("invalid rvinfo format: outer not array")
+	}
+
+	out := make([][]protocol.RvInstruction, 0, len(outer))
```
runcom, line 167: the parsing below may belong to the protocol package in go-fdo @ben-krieger

internal/db/db.go
runcom: actually nope, this belongs to the server's api anyway

internal/db/db.go
```diff
@@ -161,64 +147,373 @@ func InsertVoucher(voucher Voucher) error {
...
+		}
+		if item.Protocol != "" {
+			tp, err := transportProtocolFromString(item.Protocol)
+			if err != nil {
+				return nil, err
+			}
+			proto = tp
+		}
+
+		out = append(out, protocol.RvTO2Addr{
```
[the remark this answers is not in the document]
runcom: yep, added

internal/db/db.go
```diff
@@ -161,64 +147,373 @@ func InsertVoucher(voucher Voucher) error {
...
+	for _, item := range items {
+		var (
+			ipPtr  *net.IP
+			dnsPtr *string
+			port   uint16
+			proto  protocol.TransportProtocol
+		)
+
+		if item.IP != "" {
+			ip := net.ParseIP(item.IP)
```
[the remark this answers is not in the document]
runcom: yes, but it's ok as it's not mandated:

> A given RVTO2Addr MUST have a non-null value for at least one of RVIP and RVDNS. If both are present, the Device MUST act on the RVIP value after all RVDNS values have been attempted.

So, as long as we at least one either, we're ok - I've added that check, it's the device then that acts on it

internal/db/db.go
```diff
@@ -161,64 +148,431 @@ func InsertVoucher(voucher Voucher) error {
...
+	for i, item := range items {
+		var (
+			ipPtr  *net.IP
+			dnsPtr *string
+			port   uint16
+			proto  protocol.TransportProtocol
+		)
+
+		if item.IP != "" {
+			ip := net.ParseIP(item.IP)
```
sarmahaj, line 525: check for error and return error , if invalid IP then ParseIP will return nil and we might end up saving nil IP address in database.

api/handlers/rvinfo.go
```diff
@@ -5,158 +5,114 @@ package handlers
...
+		http.Error(w, "Error processing rvInfo", http.StatusInternalServerError)
 		return
 	}
 
-	if err := db.InsertData(rvData, "rvinfo"); err != nil {
-		slog.Debug("Error inserting rvData", "error", err)
-		http.Error(w, "Error inserting rvData", http.StatusInternalServerError)
+	if err := db.InsertRvInfo(rvInfo); err != nil {
+		slog.Debug("Error inserting rvInfo", "error", err)
+		http.Error(w, "Error inserting rvInfo", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
runcom: let's follow up with an issue but it's a good point indeed

sarmahaj (approved):
LGTM

PR #58, opened 2025-09-05
go fdo server rpms

mmartinv:
The changes in this PR add the required bits to allow the creation of an RPM file: 
- add sysusers.d files: Each service will run with a dedicated user `go-fdo-server-{manufacturer,rendezvous,owner}` and all the users will belong to the `go-fdo-server` group so they can share access to the required certificate files.
- add tmpfiles.d configuration to manage the creation of the default configuration directory `/etc/go-fdo-server`.
- add sysconfig files: to define the environment variables required by the systemd units until support for configuration files is implemented.
- add systemd unit files for each role.
- add the RPM spec file
- add Makefile targets to build the rpm/srpm and some that will be used for packit integration. 

Lot of changes to the testing workflows:
- Organize scripts and files according to https://github.com/golang-standards/project-layout
- Move shared functions to `scripts/cert-utils.sh` and `scripts/fdo-utils.sh` that will be used by the testing scripts and also by the systemd units
- Remove the `hoverkraft-tech/compose-action@v2.3.0` github action and use docker compose commands intead.
- Now it's possible to run the tests directly with:
  - `bash -c "source ./test/test-makefile.sh; test_all";`
  - `bash -c "source ./test/test-container.sh; test_all";`
- Test RPM and SRPM creation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,3 +123,32 @@ jobs:
         run: |
           source test/test-container.sh
           cleanup
+
+  test-rpms:
+    name: build srpm and rpms
+    runs-on: ubuntu-latest
+    container: fedora:latest
+    steps:
+      - name: Install dependencies
+        run: dnf install -y git make node
+
+      - name: Install golang
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.25"
+
+      - name: Check out repository code
+        uses: actions/checkout@v4
+
+      - name: Build SRPM
+        run: |
+          git config --global --add safe.directory /__w/go-fdo-server/go-fdo-server
+          make srpm
+
+      - name: Build RPM
+        run: |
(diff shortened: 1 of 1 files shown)


deployments/compose/servers.yaml
```diff
@@ -16,12 +16,12 @@ services:
      - --debug
      - manufacturing
      - manufacturer:8038
-     - --manufacturing-key=/tmp/certs/manufacturer.key
-     - --owner-cert=/tmp/certs/owner.crt
-     - --device-ca-cert=/tmp/certs/device-ca.crt
-     - --device-ca-key=/tmp/certs/device-ca.key
+     - --manufacturing-key=/tmp/go-fdo/certs/manufacturer.key
```
kgiusti, line 19: line 15 above seems to have a stray '\' char.  Not part of this patch but just noticed it.

PR #60, opened 2025-09-10
Configures .packit.yaml for Fedora and CentOS Stream builds

djach7:
Creates and configures .packit.yaml to enable Fedora and CentOS Stream Copr builds.


mmartinv:
@djach7 could you please squash all the commits down to only one and make sure the commit passes the commitlint check? e.g.: `feat: configures .packit.yaml for Fedora and CentOS Stream builds`

The rpm builds succeeded but it looks like the status is not updated here for some reason.

djach7:
Yes can do, sorry was in a meeting

PR #62, opened 2025-09-15
doc: Add user guide for installing rpms

sarmahaj:
This PR is to provide a README.md document for users to install the fdo server rpms on fedora and centos.
As a result of work created in #58 and #60 .
Keeping in draft till #60 is merged and rpms are available to be installed.


build/package/rpm/README.md
```diff
@@ -0,0 +1,65 @@
...
+LISTEN_PORT="8038"
+DATABASE_DSN="/var/lib/go-fdo-server-manufacturer/db.sqlite"
+DATABASE_PASSWORD="@PassW0rD$"
+MANUFACTURER_KEY="/etc/go-fdo-server/manufacturer.key"
+OWNER_CRT="/etc/go-fdo-server/owner.crt"
+DEVICE_CA_CRT="/etc/go-fdo-server/device-ca.crt"
+DEVICE_CA_KEY="/etc/go-fdo-server/device-ca.key"
+```
+
+If not specified in the configuration file, **certificates** are automatically generated in /etc/go-fdo-server/ when the service starts.
```
kgiusti, line 40: Double checked the code and it turns out that the certificates will be automatically generated if the files do not *exist* when the service starts. And it turns out you *always* have to specify the certificate paths in the env file else the service will fail to start.

So I'd recommend clarifying the text a bit by changing:
`If not specified in the configuration file, certificates are automatically...` 
to something like 
`If the certificate files do not exist then they are automatically...`

PR #71, opened 2025-09-19
docs: add user guide for using FSIMs

djach7:
Adds a user guide detailing usage instructions for each FSIM, with examples.

### DIFF
--- a/FSIM_USAGE.md
+++ b/FSIM_USAGE.md
@@ -1,0 +1,237 @@
+# FIDO Service Info Module (FSIM) Usage Guide
+
+This guide explains how to use the standard FIDO Service Info Modules (FSIMs) with the go-fdo-server owner service. FSIMs enable device onboarding with automated file transfers and command execution during the TO2 protocol phase.
+
+## Overview
+
+The go-fdo-server owner service supports four standard FSIMs:
+
+- **fdo.command** - Execute commands on the device
+- **fdo.download** - Download files from owner to device  
+- **fdo.upload** - Upload files from device to owner
+- **fdo.wget** - Have device download files from URLs
+
+FSIMs are activated by adding the corresponding command-line flags when starting the owner service.
+
+## Prerequisites
+
+- FDO server setup completed (see main README.md)
+- Owner service configured with proper certificates and keys
+- Device successfully initialized and voucher transferred to owner
+
+## fdo.command FSIM
+
+### Purpose
+Execute shell commands on the device during onboarding.
+
+### Usage
+```bash
+go-fdo-server owner 127.0.0.1:8043 \
+  --db /tmp/fdo/db/own.db --db-pass "$DB_PASS" \
+  --device-ca-cert /tmp/fdo/keys/device_ca_cert.pem \
+  --owner-key /tmp/fdo/keys/owner_key.der \
+  --command-date
+```
+
+### Available Commands
+**As of now only one command is available. This is for testing purposes, and the list of commands will be expanded once testing is complete.**
+- `--command-date`: Executes `date --utc` on the device to display current UTC time
+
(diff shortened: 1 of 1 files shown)


FSIM_USAGE.md
```diff
@@ -0,0 +1,170 @@
...
+  --owner-key /tmp/fdo/keys/owner_key.der \
+  --upload-directory /tmp/fdo-uploads \
+  --command-upload /var/log/device.log \
+  --command-upload /etc/device-id.txt
+```
+
+## fdo.wget FSIM
+
+### Purpose
+Instruct the device to download files from external URLs during onboarding.
```
kgiusti, line 119: Probably should add some text explaining that the device will open a new TCP connection to the HTTP server over which it will download the files rather than using the encrypted connection established between device/owner by FIDO.

The important point is that the user needs to understand the security ramifications involved.

And the go-fdo-client needs to have the --wget-dir <dir> option provided in order for this all to work. <dir> is the path to a directory on the device where the files will be downloaded to.  They retain the filenames given in the --command-wget argument.

FSIM_USAGE.md
```diff
@@ -0,0 +1,170 @@
...
+## Prerequisites
+
+- FDO server setup completed (see main README.md)
+- Owner service configured with proper certificates and keys
+- Device successfully initialized and voucher transferred to owner
+
+## fdo.command FSIM
+
+### Purpose
+Execute shell commands on the device during onboarding.
```
kgiusti, line 25: On the device side in order for this to work the command line option --echo-commands needs to be provided:

`go-fdo-client ... --echo-commands`

FSIM_USAGE.md
```diff
@@ -0,0 +1,170 @@
...
+go-fdo-server owner 127.0.0.1:8043 \
+  --db /tmp/fdo/db/own.db --db-pass "$DB_PASS" \
+  --device-ca-cert /tmp/fdo/keys/device_ca_cert.pem \
+  --owner-key /tmp/fdo/keys/owner_key.der \
+  --command-download /path/to/local/file1.txt \
+  --command-download /path/to/local/file2.conf
+```
+
+### Parameters
+- `--command-download <file_path>`: Specify a local file path to transfer to the device
```
kgiusti, line 58: On the device side the --download command line argument must be given in order for this fsim to run. --download take the path to a directory where the files will be downloaded to. The downloaded files are named using the same name given in the --command-download filepath, eg "file1.txt", "file2.conf" as above.

FSIM_USAGE.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+### Error Handling
+- If a specified file cannot be opened, the owner service will log a fatal error
+- File permissions on the owner server must allow reading by the service process
+
+## fdo.upload FSIM
+
+### Purpose
+Transfer files from the device to the owner server during onboarding.
```
kgiusti, line 83: On the device side the --upload *list* parameter must be provided to the go-fdo-client in order to trigger the file upload.  *list* is a list of directories that contain the files that can be uploaded IIUC.

FSIM_USAGE.md
```diff
@@ -0,0 +1,236 @@
...
+  --command-upload system-info.json
+```
+
+### Parameters
+- `--upload-directory <dir_path>`: Directory on owner server where uploaded files will be stored
+- `--command-upload <filename>`: Name of file to request from device
+- Upload flag can be used multiple times for multiple files
+
+### Device-Side Requirements
+The device must specify directories containing uploadable files using the `--upload` parameter:
```
say-paul, line 129: the directory path provided via --upload has is treated as relative path

djach7: Updated

PR #67, opened 2025-09-18
test(fsim-wget): add test for the wget FSIM functionality

kgiusti:
Closes: #64

PR #72, opened 2025-09-19
An alternative approach that adds multiple tests, including error scenarios

kgiusti:
This is an alternative to PR #67 

The difference with #67 is that this PR adds test for error cases.

The error tests require running onboarding twice - the first attempt fails, the second attempt proves the onboard recovers.

This requires the restarting of the owner server in one testcase.

I've also broken out the tests from the certificate generation/install of client and server so each test avoids rebuilding/reinstalling the test infrastructure.

The downside is that only the non-container case is handled at this point. We'd need to add more abstraction to allow individual control of container based servers, including the ability to restart them individually.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,3 +152,32 @@ jobs:
 
       - name: Fail if the source code has changes after build
         uses: NathanielHill/fail-if-changes@master
+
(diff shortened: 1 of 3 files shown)


kgiusti:
Purely POC, now out of date - closing.

PR #81, opened 2025-10-03
ci: add wget fsim test for missing http server

kgiusti:


test/ci/test-fsim-wget.sh
```diff
@@ -88,15 +86,66 @@ run_test () {
...
+  if [ "${rc}" == "0" ]; then
+    echo "❌ Expected device onboard to fail!"
+    return 1
+  fi
+
+  # verify that the wget FSIM error is logged
+  if ! grep -q "error handling device service info .*fdo\.wget:error" "${onboarding_log}" ; then
+    echo "❌ Client failed to report FSIM error!"
+    return 1
+  fi
```
mmartinv, line 126: As we already search in logs to check if the onboarding went well, would it make sense to write a function let's say `find_in_log_or_fail $log $string_to_find`?

mmartinv (commented):
I am not sure it does worth it to make two different functions (`fsim_wget_test` and `fsim_wget_nohost_test` and call the setup twice. I think once the `verify_equal_files` has ended it would be enough to run a new device initialization, stop the http server and check the onboarding fails. I mean we don't need to regenerate the payload files again, do we?

Apart from that it looks good to me overall just some small nits.

test/ci/test-fsim-wget.sh
kgiusti: Sorry - missed that. I'll update the PR...

test/ci/test-fsim-wget.sh
```diff
@@ -68,28 +68,70 @@ run_test () {
...
+
+  # Verify that Device 2 can successfully onboard once the HTTP server is available
+  echo "⭐ Restarting HTTP Server"
+  start_service "${wget_httpd_service_name}"
 
-  echo "⭐ Running FIDO Device Onboard with FSIM fdo.wget"
+  echo "⭐ Re-running FIDO Device Onboard with FSIM fdo.wget"
+  rm -f "${onboarding_log}"
+  rm -f "${wget_device_test_file}"
```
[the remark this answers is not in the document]
kgiusti: Oh, no - that's the destination file, the one written by the device during wget.  I'll use a better name there because it is confusing.

test/ci/test-fsim-wget.sh
```diff
@@ -42,13 +42,15 @@ start_service_owner() {
...
     --device-ca-cert="${device_ca_crt}" \
-    --command-wget "${wget_test_url}"
+    --command-wget "${wget_source_url}"
 }
 
 run_test () {
   # Add the wget_httpd service defined above
   services+=("${wget_httpd_service_name}")
 
+  trap cleanup EXIT
```
mmartinv, line 52: What's the reason of moving this line here?.

Currently it works this way:
- First we set `stop_servers` to be run before exiting. This way if the test fails we still have the logs available in `/tmp/go-fdo` and also allow us to run the test again.
- Once the tests have passed there's no need to keep `/tmp/go-fdo` so we set `cleanup` to be run before exiting.

test/ci/test-fsim-wget.sh
```diff
@@ -68,35 +70,68 @@ run_test () {
...
   run_to0 ${owner_url} "${guid}" > /dev/null
 
-  echo "⭐ Prepare the wget payload on server side: '${wget_test_file}'"
-  prepare_payload "${wget_test_file}"
+  echo "⭐ Running FIDO Device Onboard for Device 1 with FSIM fdo.wget"
+  run_fido_device_onboard --debug --wget-dir "${wget_device_download_dir}"
+
+  echo "⭐ Verify downloaded file: server: ${wget_source_file} device: ${wget_device_output_file}"
+  verify_equal_files "${wget_source_file}" "${wget_device_output_file}"
+  rm -f "${wget_device_output_file}"  # clean up result before next onboarding test
```
mmartinv, line 99: I am thinking that instead of removing the output file here we could set different output directories per device. This way we don't need to remove anything. WDYT?

test/ci/test-fsim-wget.sh
kgiusti: Ah - ok that makes sense. I moved it because the early return statement I added short-circuited the cleanup prior to registering the TRAP but as you point out we really want to keep those logs around in that case.  I'll revert this.

kgiusti:
@mmartinv looks like there's a race (?) after restarting the http server - the container tests were failing the second onboarding attempt after restarting with "connection refused" errors so I added a call to wait for the service to be ready (plus a fix to the healthcheck url in the container file but that seems irrelevant to this problem).  LMKWYT

PR #84, opened 2025-10-07
Add comprehensive OV verification for owner server

sarmahaj:
This PR: 
- prevents owner server from storing invalid/corrupted/fake ov into the databse
- Unit tests and e2e CI tests are added

### DIFF
--- a/api/handlers/vouchers.go
+++ b/api/handlers/vouchers.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"slices"
 	"time"
 
 	"github.com/fido-device-onboard/go-fdo"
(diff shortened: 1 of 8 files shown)


api/handlers/vouchers.go
```diff
@@ -110,6 +110,20 @@ func InsertVoucherHandler(ownerPKeys []crypto.PublicKey) http.HandlerFunc {
...
+				return
+			}
+
+			// Verify device certificate chain hash matches header
+			if err := ov.VerifyCertChainHash(); err != nil {
+				slog.Debug("Device cert chain hash verification failed", "guid", ov.Header.Val.GUID[:], "err", err)
+				http.Error(w, "Invalid device certificate chain hash", http.StatusBadRequest)
+				return
+			}
+
```
runcom, line 126: uhm, I think while this is useful to have (?), what we need to have is a way to tell, on insertion, that this OV belongs to us as owner of the OV itself. This doesn't really check that does it?

runcom (changes requested):

api/handlers/vouchers.go
sarmahaj, line 136: Probably the existing check here can be moved before the above two checks which i added, so that if the owners public key does not match then we can immediately discard. 
Is this the verification @runcom you were refering to ?

api/handlers/vouchers.go
```diff
@@ -110,61 +198,24 @@ func InsertVoucherHandler(ownerPKeys []crypto.PublicKey) http.HandlerFunc {
 				return
 			}
 
+			// Ov Verification
+			if err := VerifyOwnershipVoucher(&ov, ownerPKeys); err != nil {
+				slog.Debug("Ownership voucher verification failed", "guid", ov.Header.Val.GUID[:], "err", err)
```
mmartinv: Why logging it in `Debug` level when it is an `Error`?

api/handlers/vouchers.go
```diff
@@ -89,6 +94,89 @@ func GetVoucherByGUIDHandler(w http.ResponseWriter, r *http.Request) {
...
+			ownerKeyMatched = true
+			break
+		}
+	}
+	if !ownerKeyMatched {
+		return fmt.Errorf("voucher owner key does not match any of the server's configured keys")
+	}
+
+	// Header Field Validation
+	if ov.Version != FDOProtocolVersion {
```
mmartinv: The version check I was thinking of was that `ov.Version` must be the same as `ov.Header.Val.Version` but the code above opens a good question: Is the protocol version verified for all the received messages by the server? Do we support more than one protocol version? (I don't think so). Anyway the verification related to the ownership voucher and the protocol version would be the one I mentioned, checking if the ownership voucher protocol version is supported by the server should go elsewhere. Let's do a small investigation for this

api/handlers/vouchers.go
```diff
@@ -89,6 +94,89 @@ func GetVoucherByGUIDHandler(w http.ResponseWriter, r *http.Request) {
...
+			return fmt.Errorf("owner key type does not support equality comparison")
+		}
+		if eq.Equal(expectedPubKey) {
+			ownerKeyMatched = true
+			break
+		}
+	}
+	if !ownerKeyMatched {
+		return fmt.Errorf("voucher owner key does not match any of the server's configured keys")
+	}
```
mmartinv: Let's simplify this a bit and comment not obvious things, the cast is not supposed to fail because `crypto.PublicKey` implements the interface (you will need to import `slices` btw):

```suggestion
	// Cast is needed to call Equal()
	// See: https://pkg.go.dev/crypto#PublicKey
	if !slices.ContainsFunc(ownerPKeys, expectedPubKey.(interface{ Equal(crypto.PublicKey) bool }).Equal) {
		return fmt.Errorf("voucher owner key does not match any of the server's configured keys")
	}
```

test/ci/test-ov-verification.sh
```diff
@@ -0,0 +1,142 @@
...
+new_owner_url="http://${new_owner_service}"
+# needed for 'wait_for_services_ready' do not remove
+#shellcheck disable=SC2034
+new_owner_health_url="${new_owner_url}/health"
+
+start_service_new_owner () {
+  run_go_fdo_server owner ${new_owner_service} new_owner ${new_owner_pid_file} ${new_owner_log} \
+    --owner-key="${new_owner_key}" \
+    --device-ca-cert="${device_ca_crt}"
+}
```
mmartinv: Source `test-resale.sh` so you don't need to add all the definitions again:
```suggestion
source "$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )/test-resale.sh"
```

mmartinv (commented):
I would personally split the verification in two functions: `OwnershipVoucherBelongsToOwner` (or better name if you have it) and `VerifyOwnershipVoucher`.

api/handlers/vouchers.go
sarmahaj: Additional info:
The first check (ov.Version != FDOProtocolVersion) ensures we support this protocol version.
 The second check (ov.Version != ov.Header.Val.Version) cryptographically signed  detects tampering or corruption - if they don't match, someone modified the outer version or the voucher is corrupted.

api/handlers/vouchers.go
```diff
@@ -89,6 +90,90 @@ func GetVoucherByGUIDHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// VerifyVoucherOwnership verifies the ownership against owner-server.
```
mmartinv:
```suggestion
// VerifyVoucherOwnership verifies the ownership voucher belongs to this owner.
```

mmartinv:
> @mmartinv updated PR as per feedback, ptal! Also the main function to verify ov now split logically into two functions `VerifyOwnershipVoucher ` `-> VerifyVoucherOwnership -> VerifyVoucherIntegrity `
> 
> Still keeping VerifyOwnershipVoucher this signature same since we need to provide as call back in TO2 flow as well to check ov as a whole. https://github.com/fido-device-onboard/go-fdo-server/pull/84/files#diff-53c04359e19470a6c0940a0a7358b21e45df7c87ccf5f42373f0954b64e26b03R211 here

If you don't mind, I would rather prefer to use `VerifyVoucher` -> `VerifyVoucherOwnership` -> `VerifyOwnershipVoucher`

mmartinv (approved):
Great work! Thanks!

PR #100, opened 2025-10-30
refactor: switch to standard Go vendoring for cross-distro packaging

pcdubs:
Replace go-vendor-tools with standard Go module vendoring to support
building on both Fedora and CentOS/RHEL with a single spec file.

Changes:
- Commit vendor/ directory to git (998 vendored dependency files)
- Include vendor/ directory in Source0 tarball
- Set GO111MODULE=on and GOFLAGS=-mod=vendor for build and test phases
- Remove go-vendor-tools dependency from spec and Packit config
- Add vendor/modules.txt to %license for bundled dependency tracking
- Add testdata files (dc.bin, mfg_key.pem, ov.pem) to vendored go-fdo
  package required by //go:embed directives for build-time compilation

CI workflow updates:
- Exclude vendor/ from codespell spell-check to avoid false positives
  in third-party dependencies
- Exclude vendor/ from DevSkim security scanner to prevent flagging
  vendored code issues
- Disable Go dependency caching in setup-go action since vendored
  dependencies don't require go.sum

This approach follows the standard pattern used by major Go packages
in Fedora and ensures compatibility across distributions.


ben-krieger (commented):
Personally, I hate vendoring. Before Go modules, it was the best way (at least before dep), but now it's a huge pain. 

1. It requires devs to remember an extra step before pushing
2. It significantly slows down git clones and gets much worse over time with no great ways to fix 
3. It makes GitHub reviews much worse for humans

On the plus side, it makes building with Nix easier. 

At a bare minimum, the gitattributes must be changed so GitHub doesn't show vendored files in the diffs. If review pages start failing to load due to diff size (especially on mobile), it will be hard for me to help.

build/package/rpm/go-fdo-server.spec
```diff
@@ -14,19 +14,16 @@ Version:        0
 Release:        %autorelease -p
 Summary:        A Go implementation of the FIDO Device Onboard Specification
 
-# Generated by go-vendor-tools
 License:        Apache-2.0 AND BSD-3-Clause AND MIT
 URL:            %{gourl}
 Source0:        %{gosource}
-# Generated by go-vendor-tools
 Source1:        %{archivename}-vendor.tar.bz2
-Source2:        go-vendor-tools.toml
```
mmartinv: Although we are not going to use the `go-vendor-tools.toml` file from within the spec file, it might worth it to maintain it in the sources so it's possible to generate the vendor tarball externally with `go-vendor-tools`. ??

Makefile
```diff
@@ -35,20 +35,20 @@ SOURCE_DIR                 := $(CURDIR)/build/package/rpm
...
 VENDOR_TARBALL             := $(SOURCE_DIR)/$(VENDOR_TARBALL_FILENAME)
 $(VENDOR_TARBALL):
+	@echo "Creating vendor tarball..."
 	rm -rf vendor; \
-	command -v go_vendor_archive || sudo dnf install -y go-vendor-tools python3-tomlkit; \
-	go_vendor_archive create --config $(GO_VENDOR_TOOLS_FILE) --write-config --output $(VENDOR_TARBALL) .; \
-	rm -rf vendor;
+	go mod vendor; \
+	tar -cjf $(VENDOR_TARBALL) vendor/; \
+	rm -rf vendor
```
mmartinv: The problem is the spec file depending on `go-vendor-tools`, not its use in the Makefile. We can keep the current vendor tarball generation I believe

mmartinv (commented):
I think the only thing I am missing is the replacement of the licensing macros, maybe we can use the tools externally to generate the files and include them in the spec.

PR #101, opened 2025-11-03
add build version and use it as header for client requests

runcom:


.github/workflows/containers.yml
```diff
@@ -208,8 +202,7 @@ jobs:
           source test/container/test-fsim-wget.sh
           run_test
 
-      - name: Get manufacturer, rendezvous and owner server logs after failure
-        if: failure()
+      - name: Get manufacturer, rendezvous and owner server logs
```
runcom: I also don't understand why we don't capture all logs by default too - pretty easy, not expensive, and allows us to check stuff out on positive runs too...

test/ci/utils.sh
```diff
@@ -10,7 +10,7 @@ source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/../../sc
 base_dir="${PWD}/test/workdir"
 bin_dir="${base_dir}/bin"
 pid_dir="${base_dir}/run"
-logs_dir="${base_dir}/logs"
+logs_dir="$(mktemp -d)/logs"
```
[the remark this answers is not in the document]
runcom: actually this isn't fixing anything.... what's more the container tests do have no logs too..... I'm opening an issue

.github/workflows/containers.yml
runcom: https://github.com/fido-device-onboard/go-fdo-server/issues/109

test/ci/utils.sh
```diff
@@ -8,7 +8,7 @@ source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/../../sc
...
```
mmartinv: I think this change is not needed anymore and it will break local builds

PR #114, opened 2025-11-10
ci: add go test, fmt, vet to github actions

kgiusti:
No Sure If Want...

but it looks like our current set of github actions doesn't run these

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ WORKDIR /go/src/app
 COPY . .
 
 RUN apk add make curl gcc musl-dev
-RUN make && install -D -m 755 go-fdo-server /go/bin/
+RUN make build && install -D -m 755 go-fdo-server /go/bin/
 
 # Start a new stage
 FROM alpine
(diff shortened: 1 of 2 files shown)


mmartinv:
Every time a test runs `make` when it performs the server installation the `build` and the `test` targets are run. The `build` target depends on `tidy`, `fmt` and `vet`, so technically we are running them all the time.

However, it looks like we only check the source code didn't change after running `rpm` and `srpm` targets in `test-rpms` test. Maybe it makes sense to add an additional check, let's say `test-build'` in `.github/workflows/ci.yaml` before the `test-rpms`: Mainly duplicating, `test-rpms` but running `make` instead of `make rpm` and `make srpm`.

kgiusti:
@mmartinv any objection to changing the server install line in https://github.com/fido-device-onboard/go-fdo-server/blob/2a9b2ab73aef9eba53fbacac28c9b53cd48d9fc4/test/ci/utils.sh#L348 to
`make build && install -m 755 go-fdo-server ${bin_dir} && rm -f go-fdo-server`

PR #127, opened 2025-11-20
test: add e2e tests for RV bypass functionality

sarmahaj:
- Add comprehensive e2e test for RV bypass (skip TO1 and go for TO2 directly for on boarding the device) 
Note:
This PR is dependent on https://github.com/fido-device-onboard/go-fdo-client/pull/38,
so until it's merged, the tests in this PR are running on top of PR38.
I will update it to run on latest@main once it's merged.

### DIFF
--- a/test/ci/test-rv-bypass.sh
+++ b/test/ci/test-rv-bypass.sh
@@ -58,4 +58,7 @@ run_test() {
 }
 
 # Allow running directly
-[[ "${BASH_SOURCE[0]}" != "$0" ]] || { run_test; cleanup; }
+[[ "${BASH_SOURCE[0]}" != "$0" ]] || {
+  run_test
+  cleanup
+}
--- a/scripts/fdo-utils.sh
+++ b/scripts/fdo-utils.sh
@@ -12,20 +12,20 @@ set_rendezvous_info() {
   local manufacturer_url=$1
   local rendezvous_info_json=$2
   curl --fail --verbose --silent --insecure \
-       --request POST \
-       --header 'Content-Type: application/json' \
-       --data-raw "${rendezvous_info_json}" \
-       "${manufacturer_url}/api/v1/rvinfo"
+    --request POST \
+    --header 'Content-Type: application/json' \
+    --data-raw "${rendezvous_info_json}" \
+    "${manufacturer_url}/api/v1/rvinfo"
 }
 
 update_rendezvous_info() {
   local manufacturer_url=$1
   local rendezvous_info_json=$2
   curl --fail --verbose --silent --insecure \
-       --request PUT \
-       --header 'Content-Type: application/json' \
-       --data-raw "${rendezvous_info_json}" \
-       "${manufacturer_url}/api/v1/rvinfo"
+    --request PUT \
+    --header 'Content-Type: application/json' \
+    --data-raw "${rendezvous_info_json}" \
+    "${manufacturer_url}/api/v1/rvinfo"
 }
 
 get_owner_redirect_info() {


test/ci/test-rv-bypass.sh
```diff
@@ -0,0 +1,99 @@
...
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/utils.sh"
+
+verify_rv_bypass_behavior() {
+  local guid="$1"
+  local device_log="${logs_dir}/onboarding-device-${guid}.log"
+
+  echo "  🔍 Checking device onboard log for bypass indicators..."
+
+  # Device onboarding completed successfully
+  if grep -q "FIDO Device Onboard Complete" "${device_log}" || grep -q "Credential Reuse Protocol" "${device_log}"; then
```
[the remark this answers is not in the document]
sarmahaj: I think the existing pattern is more readable  plus all other tests use same style.
So skipping this suggestion.

djach7:
Tested this out, I think the e2e tests are good and seem to work as expected with my manual testing. The RV bypass API correctly accepts `rv_bypass: true`. However, I think I might've hit an `rv_bypass` bug. I think the manufacturer server might be bypassing the `rv_bypass` flag during initialization. Going to keep testing and see if I can find out more

deployments/compose/client/fdo-client-rv-bypass.yaml
```diff
@@ -0,0 +1,17 @@
...
+    user: "${container_user}"
+    networks:
+      - fdo
+    volumes:
+      - ${base_dir:-./test/workdir}:${container_working_dir:-/workdir}
+    restart: no
+networks:
+  fdo:
+    name: fdo
+    external: true
```
mmartinv: This file is not needed, it is exactly the same as the existing fdo-client.yaml, use the original one

deployments/compose/server/fdo-rv-bypass-servers.yaml
```diff
@@ -0,0 +1,66 @@
+include:
+  - fdo-networks.yaml
+services:
+  manufacturer:
+    container_name: manufacturer
+    hostname: manufacturer
+    image: manufacturer
```
mmartinv:
```suggestion
    image: go-fdo-server
```

internal/to0/to0.go
```diff
@@ -69,5 +77,13 @@ func RegisterRvBlob(rvInfo [][]protocol.RvInstruction, to0Guid string, voucherSt
...
 	}
-	return 0, fmt.Errorf("unable to register any 'RVTO2Addr' URL for guid='%x'", guid)
+
+	// If all directives were bypass, that's success (no registration needed)
+	if allBypass {
+		slog.Debug("all RV directives are bypass - no TO0 registration needed")
+		return 0, nil
+	}
+
+	// Had non-bypass directives but all failed
```
mmartinv: I think this shouldn't be done here but in `protocol.ParseOwnerRvInfo()` from the go-fdo library. According to the `Table ‑. RendezvousInfo Variables` from [FIDO 1.1 specification](https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#RVInfo), `RVBypass` applies to the Device only and shouldn't be returned as a directive to the Owner server, the same applies to many other variables. 

I am OK with getting this in the meantime but I think we should file a bug upstream and open the corresponding PR

mmartinv (changes requested):
Just some minor changes :)

internal/to0/to0.go
sarmahaj: Agreed
https://github.com/fido-device-onboard/go-fdo-server/issues/166  Created this issue to track this intended work.

kgiusti (approved):
LGTM - verified the new bypass test

Had a nit (see inline) but that can be fixed in a follow up if you want.

test/ci/utils.sh
```diff
@@ -442,20 +442,23 @@ generate_https_certs() {
...
 
-  local real_rendezvous_ip
-  real_rendezvous_ip="$(get_real_ip "${rendezvous_service_name}")"
   log_info "Checking if 'RendezvousInfo' is configured on manufacturer side (${manufacturer_url})"
   if [ -z "$(get_rendezvous_info "${manufacturer_url}")" ]; then
     log_warn "'RendezvousInfo' not found, creating it"
-    set_rendezvous_info "${manufacturer_url}" "${rendezvous_dns}" "${real_rendezvous_ip}" "${rendezvous_port}" "${rendezvous_protocol}"
+    curl --fail --verbose --silent --insecure \
+         --request POST \
+         --header 'Content-Type: text/plain' \
```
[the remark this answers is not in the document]
sarmahaj: thanks for catching that

test/ci/utils.sh
```diff
@@ -442,20 +442,23 @@ generate_https_certs() {
...
-  real_rendezvous_ip="$(get_real_ip "${rendezvous_service_name}")"
   log_info "Checking if 'RendezvousInfo' is configured on manufacturer side (${manufacturer_url})"
   if [ -z "$(get_rendezvous_info "${manufacturer_url}")" ]; then
     log_warn "'RendezvousInfo' not found, creating it"
-    set_rendezvous_info "${manufacturer_url}" "${rendezvous_dns}" "${real_rendezvous_ip}" "${rendezvous_port}" "${rendezvous_protocol}"
+    curl --fail --verbose --silent --insecure \
+         --request POST \
+         --header 'Content-Type: application/json' \
+         --data-raw "${rendezvous_info_json}" \
+         "${manufacturer_url}/api/v1/rvinfo"
```
mmartinv:
```suggestion
   set_rendezvous_info "${manufacturer_url}" "${rendezvous_info_json}"
```
Update the original function instead of replacing it with a curl command.

test/ci/utils.sh
```diff
@@ -442,20 +442,23 @@ generate_https_certs() {
...
   else
     log_info "'RendezvousInfo' found, updating it"
-    update_rendezvous_info "${manufacturer_url}" "${rendezvous_dns}" "${real_rendezvous_ip}" "${rendezvous_port}" "${rendezvous_protocol}"
+    curl --fail --verbose --silent --insecure \
+         --request PUT \
+         --header 'Content-Type: application/json' \
+         --data-raw "${rendezvous_info_json}" \
+         "${manufacturer_url}/api/v1/rvinfo"
```
mmartinv:
```suggestion
   update_rendezvous_info "${manufacturer_url}" "${rendezvous_info_json}"
```
Update the original function instead of replacing it with a curl command.

mmartinv (commented):
Please, address the pending comments and let's merge it. The failing test has nothing to do with the changes in this commit.

kgiusti:
Hi @sarmahaj - I think we need to update the test/rpm/test-onboarding-https.sh to have it set up the `rv_info` value so it can pick up the value of the rendezvous_protocol variable set by the test: https://github.com/fido-device-onboard/go-fdo-server/blob/dd5684b373bf68f44682c7fa877c5badeb204571/test/rpm/test-onboarding-https.sh#L12

That should get CI passing again (I think)!

PR #130, opened 2025-11-21
feat: add JSON API for FDO Owner Server with OpenAPI specification

djach7:
This pull request adds a JSON API for the FDO Owner Server that returns structured JSON documents while maintaining full backward compatibility with existing PEM-based clients.

  Changes

  - OpenAPI 3.0.3 Specification - Complete API specification with type-safe Go code generation
  - oapi-codegen Migration - Moved from npm-based openapi-generator to Go-native oapi-codegen v2.5.1
  - JSON Response System - All endpoints return structured JSON with proper content negotiation
  - Error Handling Consolidation - Centralized error patterns across API handlers
  - Content-Type Enforcement - Strict PEM upload validation per OpenAPI specification
  - Go Generate Workflow - Proper tool dependency management and code generation pipeline
  - Production Optimization - Eliminated debugging artifacts and fixed code smells
  - Shared Component Structure - Organized OpenAPI specs for future manufacturer/rendezvous servers

  API Endpoints Enhanced

  - /health - Server health status with version information (JSON)
  - /api/v1/ownerinfo - Owner redirect configuration management (JSON)
  - /api/v1/owner/vouchers - Ownership voucher operations (JSON)

  New Files

  - api/openapi/owner-server.yaml - OpenAPI 3.0.3 specification
  - api/openapi/generated.go - Generated Go types and server interfaces
  - api/openapi/config.yaml - oapi-codegen configuration
  - api/openapi/generate.go - go:generate implementation
  - api/openapi/README.md - Code generation documentation
  - api/openapi/shared/ - Shared components for future server specs
  - api/handlers/responses.go - JSON response utilities and error handling
  - api/handlersTest/json_api_test.go - JSON API validation tests
  - api/handlersTest/helpers_test.go - Test helper functions

  Modified Files

  - api/handlers/vouchers.go - JSON responses and simplified PEM handling
  - api/handlers/health.go - JSON health endpoint implementation
  - api/handlers/ownerinfo.go - JSON owner redirect API with error consolidation
  - api/handlers/rvinfo.go - Consistent error handling patterns
  - api/handlersTest/vouchers_test.go - Content-type validation tests
  - api/routes.go - Updated imports for generated types
  - scripts/fdo-utils.sh - Optimized utility functions for JSON API
  - Makefile - go generate workflow and tool management
  - go.mod / go.sum - oapi-codegen dependencies
  - .gitignore - Generated file handling

### DIFF
--- a/configs/goapi-codegen/components.yaml
+++ b/configs/goapi-codegen/components.yaml
@@ -1,5 +1,5 @@
 package: components
-output: internal/handlers/components/gen.go
+output: handlers/components/gen.go
 generate:
   strict-server: true
   models: true
--- a/configs/goapi-codegen/health.yaml
+++ b/configs/goapi-codegen/health.yaml
@@ -1,5 +1,5 @@
 package: health
-output: internal/handlers/health/gen.go
+output: handlers/health/gen.go
 generate:
   strict-server: true
   models: true
--- a/configs/goapi-codegen/resell.yaml
+++ b/configs/goapi-codegen/resell.yaml
@@ -1,5 +1,5 @@
 package: resell
-output: internal/handlers/resell/gen.go
+output: handlers/resell/gen.go
 generate:
   strict-server: true
   models: true
--- a/configs/goapi-codegen/rvto2addr.yaml
+++ b/configs/goapi-codegen/rvto2addr.yaml
@@ -1,5 +1,5 @@
 package: rvto2addr
-output: internal/handlers/rvto2addr/gen.go
+output: handlers/rvto2addr/gen.go
 generate:
   strict-server: true
   models: true
--- a/configs/goapi-codegen/voucher.yaml
+++ b/configs/goapi-codegen/voucher.yaml
@@ -1,5 +1,5 @@
 package: voucher
-output: internal/handlers/voucher/gen.go
+output: handlers/voucher/gen.go
 generate:
   strict-server: true
   models: true
--- a/internal/handlers/rvto2addr/impl.go
+++ b/internal/handlers/rvto2addr/impl.go
@@ -11,13 +11,23 @@ import (
 
 // Server implements RVTO2Addr HTTP handlers
 type Server struct {
-	db *db.State
+	db          *db.State
+	isOwnerMode bool // true for owner redirect info, false for rendezvous info
 }
 
-// NewServer creates a new rvto2addr server instance
+// NewServer creates a new rvto2addr server instance for rendezvous info (manufacturing server)
 func NewServer(database *db.State) *Server {
 	return &Server{
-		db: database,
+		db:          database,
+		isOwnerMode: false, // Handle rvinfo
+	}
+}
+
+// NewOwnerServer creates a new rvto2addr server instance for owner redirect info (owner server)
+func NewOwnerServer(database *db.State) *Server {
+	return &Server{
+		db:          database,
+		isOwnerMode: true, // Handle owner redirect info
 	}
 }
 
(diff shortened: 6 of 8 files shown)


Makefile
```diff
@@ -12,6 +12,53 @@ ARCH            := $(shell uname -m)
...
+	@mkdir -p $(GENERATED_DIR)/server $(GENERATED_DIR)/client
+	openapi-generator generate \
+		-i $(OPENAPI_SPEC) \
+		-g go-server \
+		-o $(GENERATED_DIR)/server \
+		--additional-properties=packageName=openapi,outputAsLibrary=true,sourceFolder=src/main/go
+	openapi-generator generate \
+		-i $(OPENAPI_SPEC) \
+		-g go \
+		-o $(GENERATED_DIR)/client \
```
mmartinv: Instead of using a Makefile target to generate the api I would prefer to use the golang native approach to generate code, please see https://github.com/oapi-codegen/oapi-codegen/?tab=readme-ov-file#for-go-124 as an example. We will need a Makefile target anyway but in this case it will run `go generate` and we will make the `build` target to depend on it.

This leads me to open the discussion about using the openapi generator tools (java) or the more native `oapi-codegen` shown in the link above.

After thinking a lot about this I think we should use the native tool as it's very well documented and used by other projects, but as I said this is open to discussion.

Makefile
```diff
@@ -12,6 +12,41 @@ VERSION         := $(shell grep 'Version:' $(SPEC_FILE) | awk '{printf "%s", $$2
...
+		echo "Error: oapi-codegen not found. Please install it first:"; \
+		echo "  go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@latest"; \
+		exit 1; \
+	}
+	@echo "OpenAPI spec validation (oapi-codegen validates during generation)"
+
+.PHONY: generate-api
+generate-api: validate-openapi
+	@echo "Generating Go types and server code from OpenAPI specification..."
+	$$(command -v oapi-codegen || echo $$(go env GOPATH)/bin/oapi-codegen) -package openapi -generate types,chi-server -o $(GENERATED_FILE) $(OPENAPI_SPEC)
```
kgiusti: It seems that the generated file is dependent on the openapi_spec file.  If that's the case then if the openapi_spec file is modified do we want to re-generate the generated code?

If that's the case I'd add GENERATED_FILE to the list of the dependencies for the build target:
```
build: $(GENERATED_FILE) tidy fmt vet
```
Then make the phony generate-api rule into a real dependency rule for GENERATED_FILE:
```
$(GENERATED_FILE): $(OPENAPI_SPEC) validate-openapi
   blab...
```
You can keep the phony generate-api rule, but it would just be a dependency on $(GENERATED_FILE) with no body
```
.PHONY: generate-api
generate-api: $(GENERATED_FILE)
```

Again this is only useful if you want "make build" to re-generate the code if the spec file has changed.

api/handlers/responses.go
```diff
@@ -0,0 +1,145 @@
...
+		http.Error(w, "Error reading body", http.StatusInternalServerError)
+		return nil, false
+	}
+	return body, true
+}
+
+// HandleDBError handles GORM errors consistently (duplicate key or not found)
+func HandleDBError(w http.ResponseWriter, r *http.Request, entityName string, err error) bool {
+	if errors.Is(err, gorm.ErrDuplicatedKey) {
+		slog.Error(entityName+" already exists (constraint)", "error", err)
```
kgiusti: WriteErrorResponse?

api/handlers/vouchers.go
```diff
@@ -58,37 +53,48 @@ func GetVoucherHandler(w http.ResponseWriter, r *http.Request) {
...
-		http.Error(w, "Invalid GUID format", http.StatusBadRequest)
+	guid, ok := ValidateAndDecodeGUID(w, r, guidHex)
+	if !ok {
 		return
 	}
 	voucher, err := db.FetchVoucher(map[string]interface{}{"guid": guid})
 	if err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
-			http.Error(w, "Voucher not found", http.StatusNotFound)
+			if ShouldReturnJSON(r) {
```
kgiusti: WriteErrorResponse?

kgiusti (commented):
Couple 'o questions:  First, why doesn't the patch use the generated ServerInterface/Handler code?  Seems like we're not leveraging a big chunk of the generated functionality.  When I look at the server examples on the oapi-codegen [website](https://github.com/oapi-codegen/oapi-codegen?tab=readme-ov-file#go-122-nethttp) it appears that we only need to provide the ServiceInterface impl and the Http server setup code.

Second, what's the reason the patch uses the chi server as opposed to the go-native (or other server options)?  Not implying it's wrong or anything - I'm not familiar with these other server impls - just interested in the motivation for choosing chi.  Thanks

.github/workflows/ci.yml
```diff
@@ -23,6 +23,9 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v4
 
+      - name: Install oapi-codegen
+        run: go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@latest
```
mmartinv: I would move this to a `Makefile` target.

.gitignore
```diff
@@ -6,3 +6,6 @@ go-fdo-server
 go-fdo-server-*.tar.*
 rpmbuild
 test/workdir
+
+# Generated OpenAPI code
+generated/
```
mmartinv: We need to track the changes in the generated code, we just cannot ignore the generated files

Makefile
```diff
@@ -12,9 +12,47 @@ VERSION         := $(shell grep 'Version:' $(SPEC_FILE) | awk '{printf "%s", $$2
...
+
+.PHONY: generate-api
+generate-api: $(GENERATED_FILE)
+
+.PHONY: clean-generated
+clean-generated:
+	@echo "Cleaning generated OpenAPI code..."
+	rm -f $(GENERATED_FILE)
```
mmartinv: This target is not needed I believe

Makefile
```diff
@@ -12,9 +12,47 @@ VERSION         := $(shell grep 'Version:' $(SPEC_FILE) | awk '{printf "%s", $$2
...
+	@command -v swagger-ui-serve >/dev/null 2>&1 || { \
+		echo "Installing swagger-ui-serve..."; \
+		npm install -g swagger-ui-serve; \
+	}
+	swagger-ui-serve $(OPENAPI_SPEC)
+
 # Build the Go project
 .PHONY: build
-build: tidy fmt vet
+build: $(GENERATED_FILE) tidy fmt vet
```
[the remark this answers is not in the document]
djach7: I implemented this in the most recent change but please take another look to make sure I've got the right idea.

test/ci/utils.sh
```diff
@@ -480,12 +481,12 @@ set_or_update_owner_redirect_info() {
...
   log_info "Checking if 'RVTO2Addr' is configured on owner side (${owner_url})"
-  if [ -z "$(get_owner_redirect_info "${owner_url}")" ]; then
+  if [ -z "$(get_ownerinfo "${owner_url}")" ]; then
     log_warn "'RVTO2Addr' not found, creating it"
-    set_owner_redirect_info "${owner_url}" "${real_owner_ip}" "${owner_dns}" "${owner_port}" "${owner_protocol}"
+    set_ownerinfo "${owner_url}" "${real_owner_ip}" "${owner_dns}" "${owner_port}" "${owner_protocol}"
   else
     log_info "'RVTO2Addr' found, updating it"
-    update_owner_redirect_info "${owner_url}" "${real_owner_ip}" "${owner_dns}" "${owner_port}" "${owner_protocol}"
+    update_ownerinfo "${owner_url}" "${real_owner_ip}" "${owner_dns}" "${owner_port}" "${owner_protocol}"
```
mmartinv: It's OK if you rename the functions but do the same in `scripts/fdo-utils.sh`: you added new functions not renamed and modified the existing ones.

.github/workflows/ci.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v4
 
+
```
mmartinv: Remove all the unwanted new lines in all the files

djach7:
Addressed in the most recent commit, I removed the paths and components from the owner spec and created a new `shared-types.go` generated file to grab those from.

api/openapi/owner-server.yaml
```diff
@@ -0,0 +1,104 @@
...
+                $ref: 'shared/schemas/common.yaml#/components/schemas/VoucherResponse'
+
+components:
+  schemas:
+    # Only owner-specific schemas - shared schemas are in shared/schemas/common.yaml
+    OwnerRedirect:
+      type: array
+      items:
+        type: object
+        required: [dns, port, protocol]
```
kgiusti: There's an "ip" entry missing here.  See https://github.com/fido-device-onboard/go-fdo-server/blob/d37e23dca1ea2d601d35911973a8c7f0f3ade513/internal/db/db.go#L511

api/openapi/generate.go
```diff
@@ -0,0 +1,6 @@
+package openapi
+
+//go:generate go tool oapi-codegen -config config.yaml owner-server.yaml
```
kgiusti: I would recommend that you try to generate the owner server code under its own owner package. In other words:
```
oapi-codegen -config config.yaml -package oapi_owner owner-server.yaml
```

We will need this eventually anyways in order to support three different ServiceInterface interfaces, as the REST API for each server will be different. 

With that each server is allowed to define a custom Server type tailored to its needs, which it will then pass to the generated Handler() function imported from its server-specific package.

The common types should continue to use the "openapi" package name since... well, they're common.

cmd/owner.go
```diff
@@ -320,15 +320,29 @@ func serveOwner(config *OwnerServerConfig) error {
...
+	apiServer := handlers.NewServer([]crypto.PublicKey{state.ownerKey.Public()}, to2Server, state.DB)
+
+	// Create the main HTTP handler with chi router for API endpoints
+	mainHandler := http.NewServeMux()
+
+	// Register FDO protocol endpoints
+	mainHandler.Handle("POST /fdo/101/msg/{msg}", handler)
+
+	// Register ownerinfo endpoint manually (not part of OpenAPI spec but needed for TO0)
+	mainHandler.HandleFunc("/api/v1/ownerinfo", handlers.OwnerInfoHandler)
```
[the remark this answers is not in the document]
djach7: My understanding was that the endpoints server different purposes. The `ownerinfo` endpoint is required for FDO TO0, and stores a JSON blob in the `owner_info` table. The `redirect` endpoint manages owner redirect configurations and provides the new JSON API interface for consumers.

api/openapi/generate.go
djach7: I refactored to generate the owner server code in its own package, should be separated now

api/handlers/ownerinfo.go
```diff
@@ -23,86 +20,66 @@ func OwnerInfoHandler(w http.ResponseWriter, r *http.Request) {
...
 	default:
-		slog.Error("Method not allowed", "method", r.Method, "path", r.URL.Path)
-		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		WriteErrorResponse(w, r, http.StatusMethodNotAllowed, "Method not allowed", "HTTP method "+r.Method+" is not supported for this endpoint", "Method not allowed")
 	}
 }
 
-func getOwnerInfo(w http.ResponseWriter, _ *http.Request) {
-	slog.Debug("Fetching ownerInfo")
+func getOwnerInfo(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
djach7: I think you are 100% correct, the ownerinfo functions were unnecessary and removed in the recent changes. As for the second bit, @runcom would you like me to work towards folding your PR (#118) into this one? I don't mind either way, I can also create a follow up PR later to make naming adjustments too.

api/components.yaml
```diff
@@ -0,0 +1,67 @@
...
+        - rvto2addr
+      properties:
+        rvto2addr:
+          type: array
+          items:
+            type: array
+            items:
+              oneOf:
+                - type: integer
+                - type: string
```
mmartinv: This definition is wrong and seems to correspond to the rvinfo not to the rvto2addr, please see https://github.com/fido-device-onboard/go-fdo-server/pull/154/changes#diff-2e7681f1cfb4420391185bff11a9619926890b4b81b8558b59ac964c251e5628R92-R113

api/rvto2addr.yaml
```diff
@@ -0,0 +1,59 @@
+openapi: 3.0.3
+info:
+  title: FDO Server - RV TO2 Address API
+  version: 1.0.0
+  description: Rendezvous TO2 address management endpoints
+
+paths:
+  /owner/redirect:
```
mmartinv:
```suggestion
  /rvto2addr:
```

internal/db/state.go
```diff
@@ -98,6 +99,68 @@ func (s *State) Close() error {
 	return sqlDB.Close()
 }
 
+// FetchRvInfoJSON reads the rvinfo JSON from the database
```
mmartinv: I think you are mixing things here: RVInfo is not part of the owner API.

PR #158, opened 2026-01-13
ci: add a short default delay to rvinfo test configuration

kgiusti:
The go-fdo-client will delay between onboarding failures. By default this delay is around 120 seconds.

The PR modifies the tests to configure a short 10 second delay to override that long default. This shortens the time it takes to run the CI tests.


kgiusti:
Let's redo this fix on top of #127 since it makes it easier for a test to set a custom RvInfo directive

PR #165, opened 2026-01-19
doc: improve README formatting and add PATH configuration note

knecasov:
This PR improves a `README` formatting and adds a note regarding a `PATH` configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ Install the client binary:
 go install github.com/fido-device-onboard/go-fdo-client@latest
 ```
 
+**Note**: Add `export PATH=$HOME/go/bin:$PATH` to your shell configuration file to run Go binaries without the `./` prefix.
+
 Generate test keys/certs (under `/tmp/fdo/keys`):
 
 ```bash


README.md
```diff
@@ -80,7 +82,7 @@ curl -fsS http://127.0.0.1:8043/health
 
 ## Managing RV Info Data
 ### Create New RV Info Data
-Send a POST request to create new RV info data, which is stored in the Manufacturer’s database:
+Send a `POST` request to create new RV info data, which is stored in the manufacturer’s database:
```
sarmahaj: Here we are referring to "Manufacturer" as one of the FDO protocol specific entity/component in this service. So I believe it should be in Title case rather than lower case. Applies to all Owner, Rendezvous or Device throughout the document. This also aligns with how it is used in FDO spec.

we can use lower case for these entities when used as sub-command names with backticks. As they are CLI commands.

kgiusti (changes requested):
Thanks for the cleanups Klara!

You need to modify your commit message to conform to this policy: https://www.conventionalcommits.org/en/v1.0.0/#summary

So for instance instead of
`README doc: improve formatting`
It should be:
`docs: improve formatting of README.md`

Otherwise looks good.

knecasov:
> Thanks for the cleanups Klara!
> 
> You need to modify your commit message to conform to this policy: https://www.conventionalcommits.org/en/v1.0.0/#summary
> 
> So for instance instead of `README doc: improve formatting` It should be: `docs: improve formatting of README.md`
> 
> Otherwise looks good.

You are welcome.

Thank you very much for the information! I fixed the commit messages.

sarmahaj (dismissed):
LGTM!! 
just one minor nit change

sarmahaj (approved):
@knecasov Thanks for improving the doc 👍

PR #180, opened 2026-02-04
chore: bump minimum Go version to 1.25.5

pcdubs:
Go 1.25.5 fixes a number of CVEs

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/fido-device-onboard/go-fdo-server
 
-go 1.25.0
+go 1.25.5
 
 require (
 	github.com/fido-device-onboard/go-fdo v0.0.0-20251217141835-8aceb06ebe21


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Copyright 2024 Intel Corporation
 # SPDX-License-Identifier: Apache 2.0
 
-FROM golang:1.25-alpine AS builder
+FROM golang:1.25.6-alpine AS builder
```
mmartinv: Isn't `golang:1.25-alpine` the latest version available? I mean, today is the same as `golang:1.25.6-alpine` but maybe tomorrow is the same as `golang:1.25.7-alpine`, or I am missing something?

sarmahaj: so that applies to https://github.com/fido-device-onboard/go-fdo-client/pull/76 as well ,pcmiw .

mmartinv (approved):

sarmahaj (approved):

PR #193, opened 2026-02-27
feat: V2 owner APIs and optional owner certificate

mmartinv:
## Summary

- **V2 voucher API** with ownership tracking, pagination, content negotiation (JSON/PEM), and 403 responses for ownership-gated extend (accepts both PUBLIC KEY and CERTIFICATE PEM; atomic ownership check inside the transaction to prevent TOCTOU races). Import returns 201 when at least one voucher imported, 200 when all detected vouchers were skipped/failed, 400 when no valid PEM blocks found. Failed vouchers (malformed, verification failure) counted as `failed`; duplicates as `skipped`.
- **V2 device listing API** with pagination, joining voucher metadata with TO2 onboarding state
- **V2 rvto2addr API** with corrected FDO transport protocol mapping
- **V2 owner handler** replacing the V1 owner package
- **Optional owner certificate** in manufacturer config — when absent, vouchers are not auto-extended at DI
- **Ownership-aware TO0 scheduling** via `ownership_verified` column: only owned vouchers are registered with the rendezvous server
- **One-time data migration** on schema upgrade: automatically sets `ownership_verified` for existing vouchers by comparing owner keys at startup
- **POST /api/v2/vouchers/verify-ownership** endpoint for manual re-verification (key rotation, interrupted migration)
- **OpenAPI request validation middleware** for all V2 endpoints (PEM schemas use `format: binary` for kin-openapi compatibility)
- **State layer hardening**: exported `IsDuplicateError`, SQL injection fix in LIKE queries, defensive certificate slice copies, `WithContext(ctx)` on all GORM transactions
- **V1 import** now compares owner keys explicitly (defensive, matches V2 pattern)
- **CI refactoring**: split monolithic `ci.yml` into `build.yml` (build+test) and `e2e.yml` (dynamic E2E test discovery from go-fdo-ci)

## Commit structure (8 commits, bisectable)

1. `refactor: extract shared voucher verification into internal/utils`
2. `refactor: harden state layer and unify key-loading helpers`
3. `feat: add OpenAPI request validation middleware`
4. `feat: add V2 APIs with ownership tracking`
5. `feat: make owner cert optional, wire V2 APIs into servers`
6. `docs: update documentation for V2 API`
7. `docs: add design spec and implementation plan for optional owner cert`
8. `ci: split CI into build and E2E workflows`

## Test plan

- [ ] All existing tests pass (`go test ./...`)
- [ ] `make all shfmt` produces no drift at every commit
- [ ] V2 voucher import: 201 when imported, 200 when skipped/failed, 400 when no PEM blocks
- [ ] V2 voucher import counts malformed/verification failures as `failed`, duplicates as `skipped`
- [ ] V2 voucher extend returns 403 when server doesn't own the voucher
- [ ] V2 voucher extend accepts both PUBLIC KEY and CERTIFICATE PEM blocks
- [ ] V2 device listing returns paginated results with TO2 state
- [ ] Manufacturer starts without owner certificate configured
- [ ] TO0 scheduler only processes ownership-verified vouchers
- [ ] OpenAPI validation rejects malformed requests before reaching handlers
- [ ] On first startup with new schema, existing vouchers are migrated (ownership_verified set by key comparison)
- [ ] `POST /api/v2/vouchers/verify-ownership` returns correct ownership counts
- [ ] `build.yml` runs build+test on push/PR
- [ ] `e2e.yml` dynamically discovers and runs E2E tests from go-fdo-ci

Resolves #228


internal/state/rvto2addr_test.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+func containsInner(s, substr string) bool {
+	for i := 0; i <= len(s)-len(substr); i++ {
+		if s[i:i+len(substr)] == substr {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
mmartinv: No longer applicable

api/v2/owner/handler.go
```diff
@@ -0,0 +1,207 @@
...
+	resellStrictHandlerV1 := v1resell.NewStrictHandler(&resellServerV1, nil)
+	v1resell.HandlerWithOptions(resellStrictHandlerV1, v1resell.StdHTTPServerOptions{BaseRouter: mgmtAPIServeMuxV1, BaseURL: "/owner"})
+
+	// Wire Voucher API
+	voucherServerV1 := v1voucher.NewServer(o.State.Voucher, []crypto.PublicKey{o.State.OwnerKey.Signer().Public()})
+	voucherStrictHandlerV1 := v1voucher.NewStrictHandler(&voucherServerV1, nil)
+	v1voucher.HandlerWithOptions(voucherStrictHandlerV1, v1voucher.StdHTTPServerOptions{BaseRouter: mgmtAPIServeMuxV1, BaseURL: "/owner"})
+
+	mgmtHandlerV1 := middleware.RateLimitMiddleware(
+		rate.NewLimiter(2, 10), // 2 req/s, burst of 10
```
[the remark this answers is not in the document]
mmartinv: I will make this configurable in a follow up PR

internal/config/manufacturer.go
```diff
@@ -59,11 +66,11 @@ func (m *ManufacturingServerConfig) Validate() error {
 		return err
 	}
 	// Validate device CA key exists
-	if err := validateCertFile(m.DeviceCA.KeyPath, "device CA key", "this key must be shared between manufacturer and owner servers"); err != nil {
+	if err := validateCertFile(m.DeviceCA.KeyPath, "device CA key", ""); err != nil {
```
sarmahaj: minor: The debug msg seems useful, shown to operators when cert files are missing.  is there a reason to drop them?

internal/config/manufacturer.go
```diff
@@ -59,11 +66,11 @@ func (m *ManufacturingServerConfig) Validate() error {
...
 		return err
 	}
 	// Validate device CA certificate exists
-	if err := validateCertFile(m.DeviceCA.CertPath, "device CA certificate", "this certificate must be shared between manufacturer and owner servers"); err != nil {
+	if err := validateCertFile(m.DeviceCA.CertPath, "device CA certificate", ""); err != nil {
```
sarmahaj: same as above
minor: The debug msg seems useful, shown to operators when cert files are missing. is there a reason to drop them?

internal/config/manufacturer.go
```diff
@@ -76,27 +83,39 @@ func (m *ManufacturingServerConfig) Validate() error {
 }
 
 // GetManufacturerKey loads the manufacturer private key
-func (m *ManufacturingServerConfig) GetManufacturerKey() (crypto.Signer, error) {
+func (m *ManufacturingServerConfig) GetManufacturerKey() (crypto.Signer, protocol.KeyType, error) {
```
sarmahaj: nit: GetManufacturerKey() and GetDeviceCAKey() are structurally identical — could be a shared loadKeyWithType(path, name string) helper? but if in future might need separate functions to expand code , please ignore this comment.

api/v2/owner/handler.go
```diff
@@ -0,0 +1,207 @@
+package owner
```
sarmahaj, line 4: nit: Missing license header , Inconsistent with other files

api/v2/voucher/handler.go
```diff
@@ -0,0 +1,676 @@
...
+
+var _ StrictServerInterface = (*Server)(nil)
+
+// ListOwnershipVouchers lists all ownership vouchers with pagination, filtering, and sorting
+func (s *Server) ListOwnershipVouchers(
+	ctx context.Context,
+	request ListOwnershipVouchersRequestObject,
+) (ListOwnershipVouchersResponseObject, error) {
+	// Set defaults
+	limit := 20
```
sarmahaj, line 58: pls consider clamping or rejecting out-of-range values here to match the spec (limit: 1-100, offset: >= 0).

api/v2/voucher/handler.go
```diff
@@ -0,0 +1,676 @@
...
+
+// GetOwnershipVoucherByGuid retrieves a single ownership voucher by GUID
+func (s *Server) GetOwnershipVoucherByGuid(
+	ctx context.Context,
+	request GetOwnershipVoucherByGuidRequestObject,
+) (GetOwnershipVoucherByGuidResponseObject, error) {
+	// Parse GUID from hex string
+	guidBytes, err := hex.DecodeString(request.Guid)
+	if err != nil || len(guidBytes) != 16 {
+		// Invalid GUID - return 404 (no 400 in spec)
```
sarmahaj: The GetOwnershipVoucherByGuid, DeleteOwnershipVoucher, and ExtendOwnershipVoucher endpoints don't define a 400 response, so malformed GUIDs return 404. Consider adding a 400
  response to the spec, re-generating the handler, and returning 400 for invalid GUID format — helps API clients distinguish "bad input" from "not found".

api/v2/voucher/handler.go
```diff
@@ -0,0 +1,676 @@
...
+		if err := utils.VerifyVoucherOwnership(&voucher, []crypto.PublicKey{serverPubKey}); err != nil {
+			result.Failed++
+			result.Messages = append(result.Messages,
+				fmt.Sprintf("voucher at position %d with GUID %s does not belong to this server: %s", position, guid, err.Error()))
+			remaining = rest
+			continue
+		}
+
+		// Verify header fields and cryptographic integrity
+		if err := utils.VerifyOwnershipVoucher(&voucher, s.DeviceCAState.Certs()); err != nil {
```
sarmahaj: s.DeviceCAState.Certs() has no nil guard, but  If DeviceCAState is nil, this might panics. Please consider adding a nil check similar to s.OwnerKeyState does on line 177.

docs/user-guide/server-config.md
```diff
@@ -79,12 +79,14 @@ provided under the `[http]` section:
 
 ## Device Certificate Authority (CA) Configuration
 
-The Device CA configuration is under the `[device_ca]` section. This section is required for both Manufacturing and Owner servers:
+The Device CA configuration is under the `[device_ca]` section. This section is required for the Manufacturing server and optional for the Owner server:
```
kgiusti: nit: did you want to include Rendezvous here in addition to Owner?  Same for the paragraph on line 89.

internal/config/manufacturer.go
```diff
@@ -31,18 +33,23 @@ func (m ManufacturingServerConfig) String() string {
...
-			detail = fmt.Sprintf(" (configured: %s)", path)
+	if path == "" {
+		errContext := ""
+		if contextLine != "" {
+			errContext = contextLine + "\n"
 		}
-		context := ""
+		return fmt.Errorf("%s is required\n%s"+
+			"run 'generate-go-fdo-server-certs.sh' for single-host setup\n"+
+			"see CERTIFICATE_SETUP.md for multi-host deployment", name, errContext)
```
kgiusti: Yikes!  This isn't introduced by the patch, but the error message isn't correct anymore (the certificate docs have moved).

I can open a separate issue for this if you don't want to fix it as this patch.

Same with line 50.

kgiusti (dismissed):
Aside from the minor docs nit it took me awhile to figure out that the manufacturing server does not actually allow voucher importing.  The manufacturer openapi.json defines a POST operation for /api/v2/vouchers.  There's a comment in manufacturer handler.go that explains that POST will fail (500?) if the key is not provided, so we can't actually POST.  Technically it should always return a 405 Method Not Allowed.

This isn't a big deal, I won't hold up the patch - it's really quite a good improvement.

api/v2/voucher/voucher_test.go
```diff
@@ -0,0 +1,597 @@
...
+		t.Errorf("Expected 1 detected, got %d", result.Detected)
+	}
+	if result.Failed != 1 {
+		t.Errorf("Expected 1 failed, got %d", result.Failed)
+	}
+}
+
+// V2 API only validates device cert chain and owner public key during import.
+// Header field validation (protocol version, GUID, DeviceInfo, etc.) happens during TO2 protocol.
+func TestInsertVoucherHandler_InvalidHeaderFields(t *testing.T) {
```
sarmahaj: immediately exits the test — the 6 sub-tests below (lines 390-492) never run and CI reports SKIP (not FAIL). 
When I tried to comment t.skip all sub-tests fail because.

api/v2/rvto2addr/handler.go
```diff
@@ -0,0 +1,241 @@
...
+
+	transportProto, err := apiToTransportProtocol(addr.Protocol)
+	if err != nil {
+		return protocol.RvTO2Addr{}, err
+	}
+
+	return protocol.RvTO2Addr{
+		IPAddress:         ipAddr,
+		DNSAddress:        addr.Dns,
+		Port:              uint16(addr.Port),
```
sarmahaj, line 198: addr.Port (int) is cast to uint16 without range check — Invalid values silently truncate/wrap. Consider validating 1-65535 before the cast.

api/v2/owner/handler.go
```diff
@@ -0,0 +1,207 @@
...
+
+	mgmtHandlerV2 := middleware.RateLimitMiddleware(
+		rate.NewLimiter(2, 10), // 2 req/s, burst of 10
+		middleware.BodySizeMiddleware(10<<20, /* 10MB */
+			mgmtAPIServeMuxV2,
+		),
+	)
+
+	// Serve OpenAPI specification
+	ownerServeMux.HandleFunc("GET /api/openapi.json", func(w http.ResponseWriter, r *http.Request) {
```
sarmahaj: nit: The block that registers GET /api/openapi.json and GET /api/docs/ (Swagger UI) is identical across owner/handler.go:189-201, manufacturer/handler.go:157-169, and rendezvous/handler.go:235-247 — only the title string differs. Could be a shared helper like middleware.ServeOpenAPI(mux, title, specJSON).

api/v2/voucher/handler.go
```diff
@@ -0,0 +1,676 @@
...
+		}, nil
+	}
+
+	// Convert to API response format (JSON)
+	summaries := make([]OwnershipVoucherSummaryInfo, len(vouchers))
+	for i, v := range vouchers {
+		var fdoVoucher fdo.Voucher
+		numEntries := 0
+		protocolVersion := 0
+		if err := cbor.Unmarshal(v.CBOR, &fdoVoucher); err == nil {
```
sarmahaj: If CBOR unmarshal fails (e.g., DB corruption), the voucher silently appears in the response with protocolVersion: 0 and numEntries: 0 — no log, no indication to the operator.
  Consider adding slog.Warn in the else branch.

api/v2/voucher/handler.go
```diff
@@ -0,0 +1,676 @@
...
+			continue
+		}
+
+		parsed = append(parsed, parsedVoucher{voucher: &voucher, position: position})
+		remaining = rest
+	}
+
+	// Treat unparsable trailing data as a failed voucher
+	if len(bytes.TrimSpace(remaining)) > 0 {
+		position++
```
sarmahaj: Trailing non-PEM data increments position, which flows into result.Detected. So if an operator uploads a PEM file with 3 valid vouchers + trailing corruption (e.g., 
  truncated file transfer), the response says detected: 4 — making it look like 4 vouchers were found when only 3 were. 
Don't increment position for trailing junk. Just count it as a failure

api/v2/voucher/handler.go
```diff
@@ -0,0 +1,676 @@
...
+	if err != nil {
+		slog.Error("Failed to parse new owner public key", "error", err)
+		return ExtendOwnershipVoucher400JSONResponse{
+			BadRequestJSONResponse: components.BadRequestJSONResponse{
+				Message: "Failed to parse new owner public key",
+			},
+		}, nil
+	}
+
+	if s.OwnerKeyState == nil {
```
sarmahaj: OwnerKeyState nil check runs after body read, GUID parse, PEM decode, and public key parse (lines 416-457). Consider moving it up to avoid wasted work when OwnerKeyState is nil.

api/v2/voucher/handler.go
```diff
@@ -0,0 +1,676 @@
...
+
+	// Parse new owner public key from PEM
+	block, _ := pem.Decode(bodyBytes)
+	if block == nil {
+		return ExtendOwnershipVoucher400JSONResponse{
+			BadRequestJSONResponse: components.BadRequestJSONResponse{
+				Message: "Invalid PEM format for owner public key",
+			},
+		}, nil
+	}
```
sarmahaj, line 453: nit: Consider checking block.Type == "PUBLIC KEY" after the nil check . Currently if a user by mistake  sends a CERTIFICATE PEM instead, they get a cryptic parse error.
so adding this check will give a clearer error message.

api/v2/rvto2addr/openapi.yaml
```diff
@@ -151,34 +147,32 @@ paths:
...
+        '401':
+          $ref: ../components/openapi.yaml#/components/responses/Unauthorized
+        '403':
+          $ref: ../components/openapi.yaml#/components/responses/Forbidden
+        '413':
+          $ref: ../components/openapi.yaml#/components/responses/PayloadTooLarge
+        '500':
+          $ref: ../components/openapi.yaml#/components/responses/InternalServerError
     delete:
       operationId: DeleteRVTO2Addr
```
sarmahaj, line 161: nit : DELETE behavior is inconsistent across V2 endpoints — DeleteOwnershipVoucher returns 204 (no body) while DeleteRVTO2Addr returns 200 (with the deleted config). Both are correct but inconsistent user experience. Consider aligning to one pattern.

api/v2/voucher/voucher_test.go
```diff
@@ -0,0 +1,597 @@
+// SPDX-FileCopyrightText: (C) 2024 Intel Corporation
```
sarmahaj, line 1: ListOwnershipVouchers, GetOwnershipVoucherByGuid, DeleteOwnershipVoucher, and ExtendOwnershipVoucher have no unit tests — only ImportOwnershipVouchers is tested. Consider adding some tests for each.

sarmahaj (commented):
Hi @mmartinv I have added some comments and suggestions. 
 Without E2E tests running the full onboarding flow through V2 Owner APIs, it's hard to approve. Waiting till we sort on E2E tests.

internal/config/manufacturer.go
mmartinv: I will fix it in the next push

api/v2/rvto2addr/openapi.yaml
mmartinv: The pattern is actually consistent: 204 for individual resources (device-ca, vouchers), 200 with body for singleton configs (rvto2addr, rvinfo)

api/v2/voucher/handler.go
mmartinv: I will add the [nethttp-middleware](https://github.com/oapi-codegen/nethttp-middleware) to all the servers so all the requests and their parameters are verified before processing them.

api/v2/voucher/handler.go
mmartinv: I am going to change the behavior a bit to be consistent with device-ca API:

- Treat all the data not contained in a PEM content block as garbage and ignore it
- Treat the unparseable data content block failed (this is the content will be detected as a voucher but it's corrupted CBOR or invalid format)
- Treats non importable data (the ov is parseable and correct but it does not belong to the owner or is not signed by a trusted ca, or the verification fails... etc) as failed

@sarmahaj does it sound reasonable to you?

api/v2/voucher/handler.go
mmartinv: Yeah, I agree, I think that would be the correct behavior, thanks for the pointer :)

internal/config/manufacturer.go
mmartinv: Not really, that was I mistake I believe, thanks for pointing this out

internal/config/manufacturer.go
mmartinv: Not really, that was I mistake I believe, thanks for pointing this out

mmartinv:
> Aside from the minor docs nit it took me awhile to figure out that the manufacturing server does not actually allow voucher importing. The manufacturer openapi.json defines a POST operation for /api/v2/vouchers. There's a comment in manufacturer handler.go that explains that POST will fail (500?) if the key is not provided, so we can't actually POST. Technically it should always return a 405 Method Not Allowed.
> 
> This isn't a big deal, I won't hold up the patch - it's really quite a good improvement.

Yeah, I think we should change the behavior on manufacturer side to make the voucher API fully functional on both the manufacturer and the owner sides: Not requiring the owner public key on manufacturer side and not extending the voucher after DI by default. 

This way the vouchers will stay in the manufacturer database without any extension and we can use the voucher API to extend the vouchers no matter who owns them (the manufacturer or any other owner)

@runcom what do you think?

api/v2/voucher/handler.go
mmartinv: Done

api/v2/voucher/handler.go
mmartinv: Done

api/v2/voucher/voucher_test.go
mmartinv: Done

runcom:
> Not requiring the owner public key on manufacturer side and not extending the voucher after DI by default.
> 
> This way the vouchers will stay in the manufacturer database without any extension and we can use the voucher API to extend the vouchers no matter who owns them (the manufacturer or any other owner)
> 
> @runcom what do you think?

Not extending by default to an owner accounts for one use case for sure but leaves out the out-of-the-box usability of the server which is why _this_ implementation has it on by default. If we want no-extension-by-default that's definitely an option we can add for users but I wouldn't change the default behavior

mmartinv:
> Not extending by default to an owner accounts for one use case for sure but leaves out the out-of-the-box usability of the server which is why _this_ implementation has it on by default. If we want no-extension-by-default that's definitely an option we can add for users but I wouldn't change the default behavior

I agree, I think we could do the following:
- Owner cert provided in manufacturing config -> maintain current behavior
- Owner cert missing -> do not extend by default -> voucher extension/resell api fully functional

Does it sound reasonable?

runcom:
> I agree, I think we could do the following:
> 
>     * Owner cert provided in manufacturing config -> maintain current behavior
> 
>     * Owner cert missing -> do not extend by default -> voucher extension/resell api fully functional
> 
> 
> Does it sound reasonable?

it sounds perfect and maintains the original behavior :+1:

api/v2/voucher/handler.go
kgiusti: Doesn't the openapi.yaml example show both PUBLIC KEY and CERTIFICATE as example extend requests?

api/v2/rvto2addr/handler.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	var ipAddr *net.IP
+	if addr.Ip != nil && *addr.Ip != "" {
+		parsed := net.ParseIP(*addr.Ip)
+		if parsed == nil {
+			return protocol.RvTO2Addr{}, fmt.Errorf("invalid IP address: %s", *addr.Ip)
+		}
+		ipAddr = &parsed
+	}
```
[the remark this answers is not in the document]
mmartinv: No, The `IPAddress` field in `protocol.RvTO2Addr` is actually of type `*net.IP`: https://pkg.go.dev/github.com/fido-device-onboard/go-fdo@v0.0.0-20260316145830-3c6ea7fc24fb/protocol#RvTO2Addr

api/v2/voucher/handler.go
mmartinv: Done

api/v2/voucher/handler.go
mmartinv: Yeah I forgot that, now the API accepts both, thank you for pointing this out

api/v2/rvto2addr/handler.go
mmartinv: Fixed by adding the validation middleware

api/v2/owner/handler.go
mmartinv: Done

mmartinv:
All the V2 tests are passing, please take a look at https://github.com/fido-device-onboard/go-fdo-ci/pull/10:

To run the tests in the `go-fdo-ci` repo go to https://github.com/fido-device-onboard/go-fdo-ci/actions/workflows/e2e.yml and run the workflow manually with the `go-fdo-server-ref` set to `refs/pull/193/head` and `go-fdo-ci-ref` set to `refs/pull/10/head`:

 https://github.com/fido-device-onboard/go-fdo-ci/actions/runs/27407428805/job/81000158911

kgiusti (dismissed):

runcom (dismissed):
lgtm

kgiusti (dismissed):
The v2 OV verification cleanups look good to me and I can confirm the logs reflect the proper responses.

api/v2/voucher/handler.go
sarmahaj: this looks good , thanks for the cleanup

api/v2/rvto2addr/openapi.yaml
sarmahaj: make sense ,thanks for clarifying!

docs/user-guide/certificates.md
```diff
@@ -48,12 +48,20 @@ The `go-fdo-server` requires three different X.509 public key certificates and t
...
+- Must be created once and distributed to both servers if Device CA verification is desired
 
-**IMPORTANT**: The Device CA certificate chain must be intact across both servers for FDO to function correctly. Independent generation breaks the signing/verification chain.
+The Device CA verification behavior varies by server role:
+
+| Server | Device CA required? | Behavior when configured | Behavior when not configured |
+|--------|--------------------|--------------------------|-----------------------------|
+| Manufacturing | Yes (for DI) | Signs device certificates; verifies device cert chain on voucher import | N/A — required for device initialization |
+| Owner | No | Only vouchers signed by a trusted Device CA are accepted during import | Vouchers from any Device CA are accepted during import |
+| Rendezvous | No | Only vouchers with a trusted Device CA cert chain are accepted during TO0 | Chain integrity is verified but any root CA is accepted during TO0 |
```
sarmahaj, line 62: The Rendezvous "Behavior when not configured" column says "Chain integrity is verified but any root CA is accepted during TO0" — but As per code I think its rejected. 
Also server-config.md (line 90) says : "TO0 requests are rejected  since no trusted roots are available." 
could you please update this to match the actual behavior.

sarmahaj (dismissed):
LGTM!
@mmartinv Thanks for resolving all comments , great work !! 
E2E on the go-fdo-ci also looks good 👍 

Just a tiny comment on the doc , not a blocker.

runcom (dismissed):

docs/user-guide/certificates.md
mmartinv: The behavior is changed for V2 APIs because of #228 and rendezvous is changed to match the behavior: If the trusted device CA is not configured then all the vouchers are allowed (by the rendezvous server in TO0 or to be imported by the owner). The incorrect documentation seems to be "TO0 requests are rejected since no trusted roots are available.". I fixed the `server-config.md`, ptal.

Thanks for taking the time to review this, it's greatly appreciated!

internal/state/deviceca.go
```diff
@@ -372,11 +388,13 @@ func (s *TrustedDeviceCACertsState) LoadTrustedDeviceCAs(ctx context.Context) er
...
 		}
 
+		certs[i] = cert
 		certPool.AddCert(cert)
 	}
 
 	// Update the state with the new cert pool
 	s.mutex.Lock()
+	s.certs = certs
 	s.certPool = certPool
```
sarmahaj, line 389: When no Device CAs are configured, s.certPool is set to an empty pool instead of nil. The go-fdo library will reject non-nil even though empty value. To match the documented behavior, please add:
 if len(certs) == 0 {
     s.certPool = nil
  } else {
      s.certPool = certPool
  }
does that make sense?

docs/user-guide/certificates.md
sarmahaj: just went through code to see if the docs match code behaviour. we need a tiny check I believe. https://github.com/fido-device-onboard/go-fdo-server/pull/193/changes#r3451703872 left a comment. Thanks for the clarification.

internal/state/deviceca.go
mmartinv: Done

sarmahaj: Thanks !

sarmahaj (dismissed):
@mmartinv lgtm Thanks for the PR 👍

runcom (approved):

sarmahaj (approved):

PR #198, opened 2026-03-10
RVInfo API Implementation for Manufacturing Server

sarmahaj:
This PR

-  Implements the RVInfo management API for the Manufacturing server as defined in OpenAPI specification PR #187.
- Provides GET, PUT, and DELETE operations for managing rendezvous information configuration on the Manufacturing server.

   ### Changes

   ### 1. Database Layer (fix: align RVInfo database parser with OpenAPI specification)
   - Implements OpenAPI RvInfo format parser: array of arrays of single-key objects `[[{dns: "..."}, {protocol: "..."}]]`
   - Enforces integer-only types for ports and delay_seconds (rejects strings)
   - Updates type definitions: `user_input` (string → boolean), `ext_rv` (string → array of strings)
   - Removes deprecated `parseHumanReadableRvJSON()` parser and related tests
   - **Breaking change**: No backward compatibility with old format (clean break for RHEL 10.3)

   ### 2. API Handler (feat: implement RVInfo API handler)
   - **GET /api/rvinfo**: Returns empty array `[]` if no config exists
   - **PUT /api/rvinfo**: Upsert operation (create or update)
   - **DELETE /api/rvinfo**: Returns deleted configuration
   - Uses StrictServerInterface pattern from OpenAPI code generation
   - Registered under `/api/` with rate limiting and body size middleware

   ### 3. Tests (test: add unit tests and E2E tests)
   - **Unit tests (5 tests)**: GET empty, GET with data, DELETE not found, PUT create, PUT invalid data
   - **E2E lifecycle test**: 10-step flow testing Empty → Invalid → Created → Updated (bypass) → Updated (multi-directive) → Deleted
   - Tests RV bypass and multi-directive configurations
   - Uses utility functions from `scripts/fdo-utils.sh`
   - CI integration: Added "Test RVInfo API Lifecycle" job

   ## Impact

   - **Manufacturing server only**: Breaking change for `/api/rvinfo` API


cmd/manufacturing.go
```diff
@@ -263,24 +267,65 @@ func serveManufacturing(config *ManufacturingServerConfig) error {
 		},
 	}
 
-	// Handle messages
-	apiRouter := http.NewServeMux()
-	apiRouter.HandleFunc("GET /vouchers", handlers.GetVoucherHandler)
-	apiRouter.HandleFunc("GET /vouchers/{guid}", handlers.GetVoucherByGUIDHandler)
-	apiRouter.Handle("/rvinfo", handlers.RvInfoHandler())
```
mmartinv: You still need to wire the old handlers with `/api/v1` prefix. The new API will have `/api/v2`.

For that, you can create two servemux, one for each API version, e.g.: `mgmtAPIServeMuxV1` and `mgmtAPIServeMuxV2`

mmartinv (commented):
We still need to wire the old V1 APIs and we need to use the new `internal/state` instead of the deprecated `internal/db`. Please see review comments above 

Also, instead of modifying existing tests let's create new for version 2: `test-rv-bypass-v2.sh` and `test-rvinfo-api-v2.sh`.

For `test-rv-bypass-v2.sh`, for example, you just need to copy the old one and overwrite the `rvinfo` related functions to set the new api endpoints (and use the new `rvinfo` format)

test/ci/test-rv-bypass.sh
```diff
@@ -34,10 +34,10 @@ run_test() {
...
   wait_for_services_ready
 
-  log_info "Setting Rendezvous Info with RV BYPASS flag"
-  # Override rv_info to point to owner with rv_bypass flag
-  rv_info="[{\"dns\": \"${owner_dns}\", \"device_port\": \"${owner_port}\", \"protocol\": \"${owner_protocol}\", \"ip\": \"${owner_ip}\", \"owner_port\": \"${owner_port}\", \"rv_bypass\": true}]"
-  set_or_update_rendezvous_info "${manufacturer_url}" "${rv_info}"
+  log_info "Setting Rendezvous Info with RV BYPASS flag using V1 API"
+  # V1 format: array of arrays with string ports, rv_bypass flag
+  rv_info_v1="[[{\"dns\": \"${owner_dns}\"}, {\"device_port\": \"${owner_port}\"}, {\"protocol\": \"${owner_protocol}\"}, {\"ip\": \"${owner_ip}\"}, {\"owner_port\": \"${owner_port}\"}, {\"rv_bypass\": true}]]"
+  set_or_update_rendezvous_info "${manufacturer_url}" "${rv_info_v1}"
```
mmartinv: This test is for v1 so this change is not needed

mmartinv (commented):
Please, lets keep the old v1 implementation as it is, in this PR we are supposed to implement and wire the new v2 version only

sarmahaj:
@mmartinv 
Thanks for the last review!
I think I had made it a lil complicated earlier when we decided to have back words compatibility and thought we need some refactoring for v1 implementation as well (state layer implementation) but now made it simpler .
1. So V1 implementation is untouched and all ci tests are running by default with V1 apis. 
2. V2 implementation has a few CI tests(e2e lifecycle, onboarding and rvbypass) .
3. Also DI handler updated to take care of use case when user creates RvInfo using V2 but DI needs to check both tables for V1 and V2 in order to fetch the RVInfo. (since we have two different tables for V1 and V2)

Could you Ptal Thanks !

api/definitions/manufacturer.yaml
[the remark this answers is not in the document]
mmartinv: Please do not modify this file, you don't need to and I will modify all the structure in my PR once this one is merged

cmd/manufacturing.go
```diff
@@ -257,30 +270,90 @@ func serveManufacturing(config *ManufacturingServerConfig) error {
...
+					return rvInfo, nil
+				}
+				slog.Debug("RvInfo not found in legacy table 'rvinfo', checking V2 table 'rv_info'")
+
+				// Fallback to V2 table (new V2 API users)
+				rvInfoJSON, err := rvInfoState.FetchRvInfoJSON(ctx)
+				if err != nil {
+					return nil, err
+				}
+				return manufacturing.ParseOpenAPIRvJSON(rvInfoJSON)
```
mmartinv: I think it would be more correct to save the `RendezvousInfo` to the database in the same format, no matter what API version or persistent state implementation is used. For that I suggest to save the `[][]protocol.Instruction` in CBOR format, for deprecated `db` code I would look like this:

```go

func FetchRvInfoJSON() ([]byte, error) {
	var rvInfoRow RvInfo
	if err := db.Where("id = ?", 1).First(&rvInfoRow).Error; err != nil {
		return nil, err
	}
	var rvInfo [][]protocol.RvInstruction
    // try to decode the data within the database as CBOR
	if err := cbor.Unmarshal(rvInfoRow.Value, &rvInfo); err != nil {
        // try to decode the  data within the database as JSON
		if rvInfo, err = parseHumanReadableRvJSON(rvInfoRow.Value); err != nil {
			return nil, ErrInvalidRvInfo
		}
        // Use CBOR to encode the rvInfo 
		rvInfoCBOR, err := cbor.Marshal(rvInfo)
		if err != nil {
			return nil, err
		}
        // Update the database with the new format
		tx := db.Model(&RvInfo{}).Where("id = ?", 1).Update("value", rvInfoCBOR)
		if tx.Error != nil {
			return nil, tx.Error
		}
        // this shoudn't happen as the rvInfoRow was retrieved above.
		if tx.RowsAffected == 0 {
			return nil, gorm.ErrRecordNotFound
		}
	}
    // rvInfoToJson has to be implemented but it does just the opposite as parseHumanReadableRvJSON function
	return rvInfoToJSON(rvInfo)
}


func InsertOwnerInfo(data []byte) error {
    // We get the [][]protocol.Instruction representation submitted by the user
    rvInfo, err = parseHumanReadableRvJSON(data)
    if err != nil {
        return fmt.Errorf("%w: %v", ErrInvalidRvInfo, err)
    }

    // Instead of saving the data directly we encode it with CBOR first
    rvInfoCBOR, err := cbor.Marshal(rvInfo)
	if err != nil {
		return nil, err
	}

	rvInfo := db.RvInfo{
		ID:    1,
		Value: rvInfoCBOR,
	}

	tx := db.Clauses(clause.OnConflict{DoNothing: true}).Create(&rvInfo)
	if tx.Error != nil {
		return tx.Error
	}
	if tx.RowsAffected == 0 {
		return gorm.ErrDuplicatedKey
	}
    return nil
}

func UpdateRvInfo(data []byte) error {
	// check the data can be parsed into [][]protocol.RvInstruction
	if rvInfo, err := parseHumanReadableRvJSON(data); err != nil {
		return fmt.Errorf("%w: %v", ErrInvalidRvInfo, err)
	}

    // Instead of saving the data directly we encode it with CBOR first
    rvInfoCBOR, err := cbor.Marshal(rvInfo)
	if err != nil {
		return nil, err
	}

	tx := db.Model(&RvInfo{}).Where("id = ?", 1).Update("value", rvInfoCBOR)
	if tx.Error != nil {
		return tx.Error
	}
	if tx.RowsAffected == 0 {
		return gorm.ErrRecordNotFound
	}
	return nil
}
```

For the new implementation in `state` package it would be mostly the same but in this case the `parseHumanReadableRvJSON` and `rvInfoToJSON` implementations will be different

cmd/manufacturing.go
```diff
@@ -257,30 +270,90 @@ func serveManufacturing(config *ManufacturingServerConfig) error {
...
+	rvInfoStrictHandler := rvinfo.NewStrictHandler(&rvInfoServer, nil)
+	rvinfo.HandlerFromMux(rvInfoStrictHandler, mgmtAPIServeMuxV2)
+	// TODO: Add voucher V2 API handlers here following the same pattern (tracked in PR #193)
+	mgmtAPIHandlerV2 := rateLimitMiddleware(rate.NewLimiter(2, 10),
+		bodySizeMiddleware(1<<20, mgmtAPIServeMuxV2))
+	rootMux.Handle("/api/v2/", http.StripPrefix("/api/v2", mgmtAPIHandlerV2))
 
 	// Listen and serve
-	server := NewManufacturingServer(config.HTTP, httpHandler)
+	server := NewManufacturingServer(config.HTTP, rootMux)
```
mmartinv: All this code should go in `internal/handlers/manufacturing/handler.go`, take `internal/handlers/rendezvous/handler.go` as an example

cmd/manufacturing.go
```diff
@@ -257,30 +270,90 @@ func serveManufacturing(config *ManufacturingServerConfig) error {
...
+			io.Reader
+			io.Closer
+		}{
+			Reader: io.LimitReader(r.Body, limitBytes),
+			Closer: r.Body,
+		}
+		next.ServeHTTP(w, r)
+	}
+}
+
```
mmartinv: All this code should go in internal/handlers/manufacturing/handler.go, take internal/handlers/rendezvous/handler.go as an example

test/ci/test-onboarding-v2.sh
```diff
@@ -0,0 +1,77 @@
...
+  log_info "Unsetting the error trap handler"
+  trap - EXIT
+  test_pass
+}
+
+# Allow running directly
+[[ "${BASH_SOURCE[0]}" != "$0" ]] || {
+  run_test
+  cleanup
+}
```
mmartinv: You just need to overwrite `rv_info` with the new format and the existing functions no need to put a `v2` suffix.

Probably it would be even better to move these functions to `scripts/fdo-api-v2.sh` and do:
```bash
#! /usr/bin/env bash

set -euo pipefail

source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/test-onboarding.sh"
source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/../../scripts/fdo-api-v2.sh"

rv_info="[[{\"dns\":\"${rendezvous_dns}\"},{\"device_port\":${rendezvous_port}},{\"protocol\":\"${rendezvous_protocol}\"},{\"ip\":\"${rendezvous_ip}\"},{\"owner_port\":${rendezvous_port}}]]"

# Allow running directly
[[ "${BASH_SOURCE[0]}" != "$0" ]] || {
  run_test
  cleanup
}
```

This way we can add the rest of the overwrites there and share them with the rest of the scripts

mmartinv (commented):
I think we are getting closer to have something mergeable. Thanks @sarmahaj for your work!

internal/db/db.go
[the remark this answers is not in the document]
mmartinv: Actually, after reviewing this more deeply we will need some changes here

test/ci/test-onboarding-v2.sh
mmartinv: Oh, I forgot about the contents of `scripts/fdo-api-v2.sh`:

```bash
#! /usr/bin/env bash

set -euo pipefail

# V2 API functions (local overrides)
get_rendezvous_info() {
  local manufacturer_url=$1
  curl --fail --verbose --silent --insecure \
    --request GET \
    "${manufacturer_url}/api/v2/rvinfo"
}

set_rendezvous_info() {
  local manufacturer_url=$1
  local rv_info_v2=$2
  curl --fail --verbose --silent --insecure \
    --request PUT \
    --header 'Content-Type: application/json' \
    --data-raw "${rv_info_v2}" \
    "${manufacturer_url}/api/v2/rvinfo"
}

update_rendezvous_info() {
  local manufacturer_url=$1
  local rv_info_v2=$2
  curl --fail --verbose --silent --insecure \
    --request PUT \
    --header 'Content-Type: application/json' \
    --data-raw "${rv_info_v2}" \
    "${manufacturer_url}/api/v2/rvinfo"
}

delete_rendezvous_info() {
  local manufacturer_url=$1
  curl --fail --verbose --silent --insecure \
    --request DELETE \
    "${manufacturer_url}/api/v2/rvinfo"
}
```

test/ci/test-rv-bypass-v2.sh
```diff
@@ -0,0 +1,75 @@
...
+  log_info "Unsetting the error trap handler"
+  trap - EXIT
+  test_pass
+}
+
+# Allow running directly
+[[ "${BASH_SOURCE[0]}" != "$0" ]] || {
+  run_test
+  cleanup
+}
```
mmartinv, line 17: Following the example of `test-onboarding-2.sh`, this file would look like this:
```suggestion
#! /usr/bin/env bash

set -euo pipefail

source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/test-rv-bypass.sh"
source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/../../scripts/fdo-api-v2.sh"

rv_info="[[{\"dns\":\"${rendezvous_dns}\"},{\"device_port\":${rendezvous_port}},{\"protocol\":\"${rendezvous_protocol}\"},{\"ip\":\"${rendezvous_ip}\"},{\"owner_port\":${rendezvous_port}}]]"

# Allow running directly
[[ "${BASH_SOURCE[0]}" != "$0" ]] || {
  run_test
  cleanup
}
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
+#! /usr/bin/env bash
+
+set -euo pipefail
+
+source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/utils.sh"
```
mmartinv, line 5:
```suggestion
source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/utils.sh"
source "$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)/../../scripts/fdo-api-v2.sh"
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+    --data-raw "${rv_info_v2}" \
+    "${manufacturer_url}/api/v2/rvinfo"
+}
+
+delete_rendezvous_info_v2() {
+  local manufacturer_url=$1
+  curl --fail --verbose --silent --insecure \
+    --request DELETE \
+    "${manufacturer_url}/api/v2/rvinfo"
+}
```
mmartinv:
```suggestion
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+  start_services "${manufacturer_service_name}"
+
+  log_info "Wait for manufacturer service to be ready"
+  wait_for_services_ready "${manufacturer_service_name}"
+
+  log_info "=== Testing RVInfo API Lifecycle ==="
+
+  # Step 1: Verify empty state (baseline)
+  log_info "Step 1: GET /api/v2/rvinfo - Verify empty state"
+  response=$(get_rendezvous_info_v2 "${manufacturer_url}")
```
mmartinv:
```suggestion
  response=$(get_rendezvous_info "${manufacturer_url}")
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+  log_info "✅ Initial state is empty"
+
+  # Step 2: Test validation - invalid data (negative test)
+  log_info "Step 2: PUT /api/v2/rvinfo - Test validation (missing DNS/IP)"
+  invalid_rvinfo='[[
+    {"protocol": "http"},
+    {"owner_port": 8080}
+  ]]'
+
+  if update_rendezvous_info_v2 "${manufacturer_url}" "${invalid_rvinfo}" 2>/dev/null; then
```
mmartinv:
```suggestion
  if update_rendezvous_info "${manufacturer_url}" "${invalid_rvinfo}" 2>/dev/null; then
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+
+  # Step 3: Create RVInfo configuration
+  log_info "Step 3: PUT /api/v2/rvinfo - Create initial configuration"
+  initial_rvinfo='[[
+    {"dns": "rv.example.com"},
+    {"protocol": "https"},
+    {"owner_port": 8443}
+  ]]'
+
+  response=$(update_rendezvous_info_v2 "${manufacturer_url}" "${initial_rvinfo}")
```
mmartinv:
```suggestion
  response=$(update_rendezvous_info "${manufacturer_url}" "${initial_rvinfo}")
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+
+  if ! echo "${response}" | jq -e '.[0] | type == "array"' >/dev/null; then
+    log_error "Response is not an array of arrays"
+    return 1
+  fi
+  log_info "✅ RVInfo created successfully"
+
+  # Step 4: Verify creation
+  log_info "Step 4: GET /api/v2/rvinfo - Verify creation"
+  response=$(get_rendezvous_info_v2 "${manufacturer_url}")
```
mmartinv:
```suggestion
  response=$(get_rendezvous_info "${manufacturer_url}")
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+  # Step 5: Update with RV bypass configuration
+  log_info "Step 5: PUT /api/v2/rvinfo - Update with RV bypass"
+  bypass_rvinfo='[[
+    {"dns": "owner.example.com"},
+    {"protocol": "https"},
+    {"owner_port": 8443},
+    {"rv_bypass": true}
+  ]]'
+
+  response=$(update_rendezvous_info_v2 "${manufacturer_url}" "${bypass_rvinfo}")
```
mmartinv:
```suggestion
  response=$(update_rendezvous_info ${manufacturer_url}" "${bypass_rvinfo}")
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+  log_info "Response: ${response}"
+  log_info "✅ RVInfo updated with RV bypass"
+
+  # Step 6: Verify RV bypass configuration
+  log_info "Step 6: GET /api/v2/rvinfo - Verify RV bypass"
+  response=$(get_rendezvous_info_v2 "${manufacturer_url}")
```
mmartinv:
```suggestion
  response=$(get_rendezvous_info "${manufacturer_url}")
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+    ]
+  ]'
+
+  response=$(update_rendezvous_info_v2 "${manufacturer_url}" "${multi_rvinfo}")
+  log_info "Response: ${response}"
+  log_info "✅ RVInfo updated with multiple directives"
+
+  # Step 8: Verify multi-directive configuration
+  log_info "Step 8: GET /api/v2/rvinfo - Verify multiple directives"
+  response=$(get_rendezvous_info_v2 "${manufacturer_url}")
```
mmartinv:
```suggestion
  response=$(get_rendezvous_info "${manufacturer_url}")
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+      {"owner_port": 8443}
+    ],
+    [
+      {"ip": "192.168.1.100"},
+      {"protocol": "http"},
+      {"owner_port": 8080}
+    ]
+  ]'
+
+  response=$(update_rendezvous_info_v2 "${manufacturer_url}" "${multi_rvinfo}")
```
mmartinv:
```suggestion
  response=$(update_rendezvous_info "${manufacturer_url}" "${multi_rvinfo}")
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+  # Verify second directive has IP
+  if ! echo "${response}" | jq -e '.[1][0] | has("ip")' >/dev/null; then
+    log_error "Second directive should have IP"
+    return 1
+  fi
+  log_info "✅ Multiple directives verified (DNS + IP fallback)"
+
+  # Step 9: Delete RVInfo configuration
+  log_info "Step 9: DELETE /api/v2/rvinfo - Delete configuration"
+  response=$(delete_rendezvous_info_v2 "${manufacturer_url}")
```
mmartinv:
```suggestion
  response=$(delete_rendezvous_info "${manufacturer_url}")
```

test/ci/test-rvinfo-apis-v2.sh
```diff
@@ -0,0 +1,232 @@
...
+  # Verify response contains the deleted configuration
+  if ! echo "${response}" | jq -e 'type == "array"' >/dev/null; then
+    log_error "DELETE should return the deleted configuration"
+    return 1
+  fi
+  log_info "✅ RVInfo deleted, returned deleted configuration"
+
+  # Step 10: Verify deletion (back to empty state)
+  log_info "Step 10: GET /api/v2/rvinfo - Verify deletion"
+  response=$(get_rendezvous_info_v2 "${manufacturer_url}")
```
mmartinv:
```suggestion
  response=$(get_rendezvous_info "${manufacturer_url}")
```

cmd/manufacturing.go
sarmahaj: > I think it would be more correct to save the `RendezvousInfo` to the database in the same format, no matter what API version or persistent state implementation is used. For that I suggest to save the `[][]protocol.Instruction` in CBOR format, for deprecated `db` code I would look like this:

Absolutely agree!
Does that also imply we use a single table in database which makes more sense to have a single source of truth for RvInfo ? In contrast to what we have currently two different tables for V1 and V2? 
@mmartinv Please let me know wdyt?

mmartinv: yes, absolutely

sarmahaj: done these chanegs now ptal, Thanks !

api/definitions/manufacturer.yaml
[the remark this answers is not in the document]
sarmahaj: 1. For V2 API handlers generation from rvinfo.yaml to work correctly, the path key needed to be changed from `/v1/rvinfo` to `/rvinfo` because the version prefix (/v2/) is
  handled at the server/router level, not in the definition file itself.
If not changes it becomes something like `/api/v2/v1/rvinfo`  rather than `/api/v2/v1/rvinfo`

  2. Why manufacturer.yaml had to change:
  Once we updated the path key in rvinfo.yaml, the $ref in manufacturer.yaml must point to the correct key: 
This resolves my build errors for now.
I was planning to change some names to match v2 version but I suppose you are going to handle that part in your PR.

sarmahaj:
@mmartinv 
Thanks for the review and suggestions ptal!
 
- Now we only have single database table for both api versions
- we save rvinfo in CBOR fromat
- so related refactoring in both API versions
- some more testing added

internal/db/db.go
```diff
@@ -218,12 +232,215 @@ func UpdateRvInfo(data []byte) error {
...
+		tx := db.Model(&RvInfo{}).Where("id = ?", 1).Update("value", rvInfoCBOR)
+		if tx.Error != nil {
+			return nil, tx.Error
+		}
+		// This shouldn't happen as rvInfoRow was retrieved above
+		if tx.RowsAffected == 0 {
+			return nil, gorm.ErrRecordNotFound
+		}
+	}
+
```
mmartinv: Why duplicating `FetchRvInfo()` code, isn't it possible to call it from here?

```suggestion
    rvInfo, err := FetchRvInfo()
    if err != nil {
       return nil, err
    }
```

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/fido-device-onboard/go-fdo-server
 
-go 1.25.7
+go 1.25.8
```
mmartinv: This breaks RPM builds

cmd/manufacturing.go
```diff
@@ -257,30 +270,90 @@ func serveManufacturing(config *ManufacturingServerConfig) error {
...
+	healthStrictHandler := health.NewStrictHandler(&healthServer, nil)
+	health.HandlerFromMux(healthStrictHandler, rootMux)
+
+	// === V1 API (Old handlers for backward compatibility) ===
+	mgmtAPIServeMuxV1 := http.NewServeMux()
+	mgmtAPIServeMuxV1.HandleFunc("GET /vouchers", handlers.GetVoucherHandler)
+	mgmtAPIServeMuxV1.HandleFunc("GET /vouchers/{guid}", handlers.GetVoucherByGUIDHandler)
+	mgmtAPIServeMuxV1.Handle("/rvinfo", handlers.RvInfoHandler())
+	mgmtAPIHandlerV1 := rateLimitMiddleware(rate.NewLimiter(2, 10),
+		bodySizeMiddleware(1<<20, mgmtAPIServeMuxV1))
```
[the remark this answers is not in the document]
mmartinv: We will implement this as a follow up PR

cmd/manufacturing.go
```diff
@@ -257,30 +270,90 @@ func serveManufacturing(config *ManufacturingServerConfig) error {
...
+
+	// === V2 API (New OpenAPI handlers) ===
+
+	mgmtAPIServeMuxV2 := http.NewServeMux()
+	rvInfoServer := rvinfo.NewServer(rvInfoState)
+	rvInfoStrictHandler := rvinfo.NewStrictHandler(&rvInfoServer, nil)
+	rvinfo.HandlerFromMux(rvInfoStrictHandler, mgmtAPIServeMuxV2)
+	// TODO: Add voucher V2 API handlers here following the same pattern (tracked in PR #193)
+	mgmtAPIHandlerV2 := rateLimitMiddleware(rate.NewLimiter(2, 10),
+		bodySizeMiddleware(1<<20, mgmtAPIServeMuxV2))
```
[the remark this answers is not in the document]
mmartinv: We will implement this in a follow up PR

test/ci/test-onboarding-v2.sh
mmartinv: @sarmahaj I see you maintained the `run_test` function here, didn't my suggestion work?

test/ci/test-rv-bypass-v2.sh
mmartinv: @sarmahaj same as before, didn't this work?

cmd/manufacturing.go
mmartinv: After reviewing this again I think we are mixing API logic with persistent state logic. Let's leave this as it is and I will refactor it in a follow up PR

test/ci/test-onboarding-v2.sh
sarmahaj: Sorry I missed on this earlier , thanks for suggesting !

test/ci/test-rv-bypass-v2.sh
sarmahaj: sorry I missed on this earlier, updated now 👍

internal/db/db.go
sarmahaj: Thanks for catching this!
Removed this duplicated code and also at some other places now in `refactor: remove duplicate RvInfo` in this commit

test/ci/test-rv-bypass-v2.sh
mmartinv: Is not enough to set the `rv_info` with the new format?

sarmahaj: updated now

sarmahaj: Base test-rv-bypass.sh reassigns rv_info inside run_test(), so we must override the function

internal/state/rvinfo.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	var rvInfo [][]protocol.RvInstruction
+
+	// Try to unmarshal as CBOR first (new format)
+	if err := cbor.Unmarshal(rvInfoRow.Value, &rvInfo); err != nil {
+		// CBOR failed, try to parse as V2 OpenAPI JSON (old format)
+		// This handles migration if database still has JSON from before CBOR implementation
+		var parseErr error
+		rvInfo, parseErr = ParseOpenAPIRvJSON(rvInfoRow.Value)
```
kgiusti: I'm a bit confused here. This implies that V2 JSON has been stored to the database. But IIUC V2 will never write JSON to the database, only CBOR.  So if a V2 fetch reads JSON from the database won't it be V1 JSON, not V2 JSON?

internal/state/rvinfo.go
```diff
@@ -0,0 +1,268 @@
...
+	ErrRvInfoNotFound = errors.New("rendezvous info not found")
+)
+
+// RvInfoState manages rendezvous information configuration state
+type RvInfoState struct {
+	DB *gorm.DB
+}
+
+// RvInfo model stores rendezvous information as CBOR-encoded [][]protocol.RvInstruction
+type RvInfo struct {
```
kgiusti, line 32: This is a duplicate of the RvInfo type in internal/db/models.go.

Can one of the definitions be removed?

internal/state/rvinfo.go
```diff
@@ -0,0 +1,268 @@
...
+	ID    int    `gorm:"primaryKey;check:id = 1"`
+	Value []byte `gorm:"not null"` // CBOR-encoded [][]protocol.RvInstruction
+}
+
+// TableName specifies the table name for RvInfo model
+// Uses same table as V1 API for unified storage
+func (RvInfo) TableName() string {
```
kgiusti, line 39: Also duplicated in internal/db/models.go

internal/handlers/rvinfo/handler.go
```diff
@@ -0,0 +1,155 @@
...
+	rvInfoJSON, err := json.Marshal(request.Body)
+	if err != nil {
+		slog.Error("failed to marshal RV info", "error", err)
+		return UpdateRendezvousInfo400JSONResponse{
+			components.BadRequest{Message: "invalid rendezvous info format"},
+		}, nil
+	}
+
+	// Validate the RV info can be parsed (before storing)
+	if _, err := state.ParseOpenAPIRvJSON(rvInfoJSON); err != nil {
```
kgiusti: nit: s.RvInfo.UpdateRvInfo repeats this - calls ParseOpenAPIRvJSON to validate the JSON there as well (returns ErrInvalidRvInfo, handled below).  Not wrong, but doubled parsed.

internal/state/rvinfo_parser.go
```diff
@@ -0,0 +1,406 @@
...
+	// Validate port range (1-65535)
+	if f < 1 || f > 65535 {
+		return 0, fmt.Errorf("port must be between 1 and 65535, got %v", f)
+	}
+
+	return uint16(f), nil
+}
+
+// protocolCodeFromString converts protocol string to protocol code
+func protocolCodeFromString(s string) (uint8, error) {
```
kgiusti: Nit: This needs to be kept in sync with the ProtocolStringFromCode in utils/rvinfo.go.  It is also duplicated in internal/db/db.go. It would make sense t move this to utils/rvinfo.go with its partner function and remove it from db.go

internal/state/rvinfo_parser.go
```diff
@@ -0,0 +1,406 @@
...
+		return uint8(protocol.RVProtCoapTCP), nil
+	case "coap":
+		return uint8(protocol.RVProtCoapUDP), nil
+	default:
+		return 0, fmt.Errorf("unsupported protocol %q", s)
+	}
+}
+
+// parseMediumValue parses the medium value (network interface type)
+func parseMediumValue(v any) (uint8, error) {
```
kgiusti: Nit: same as protocolCodeFromString - has inverse function in utils/rvinfo.go and duplication in db.go

kgiusti (commented):
I had one point I need clarification on (see inline), otherwise a few non-critical code duplication nits that could be handled in a follow up PR if desired.

test/ci/test-rv-bypass-v2.sh
mmartinv: Please move the `rv_info` variable outside the `run_test` function in `test-rv-bypass.sh`

internal/state/rvinfo.go
mmartinv: Yeah, this will be the final one, the one in `db` package will be removed in #196

internal/state/rvinfo.go
mmartinv: The same as above `internal/db` is deprecated and will be removed as part of #196

internal/state/rvinfo.go
mmartinv: Yes, you are right. this should be `parseHumanReadableRvJSON(rvInfoRow.Value)` from `db` package which is a private function. We can make it public temporarily until we merge #196.

I think the correct way of doing this would be moving the JSON parsing to the corresponding API handling code and the persistence layer just saves/retrieves the CBOR representantion of `[][]protocol.RvInstruction`. 

I planned to do this in #196.

internal/handlers/rvinfo/handler.go
mmartinv: Yes, you are right again. Actually this should not discard the `[][]protocol.RvInstruction` but pass it to the `UpdateRvInfo` function which should accept `[][]protocol.RvInstruction` as an argument instead of a JSON. As I said in another comment, the JSON parsing should be API's responsability not persistence layer's

internal/state/rvinfo.go
sarmahaj: @kgiusti  absolutely right!
- here if CBOR unmarshal fails, the database would contain V1 JSON only not V2 JSON.

 @mmartinv Agreed 
For this PR: I'll update the fallback to use db.parseHumanReadableRvJSON() (making it public temporarily). Once #196 merges, we can remove this migration logic.

internal/db/db.go
```diff
@@ -296,11 +478,13 @@
...
 }
 
-// parseHumanReadableRvJSON parses a JSON like
+// ParseHumanReadableRvJSON parses a JSON like
 // [{"dns":"fdo.example.com","device_port":"8082","owner_port":"8082","protocol":"http","ip":"127.0.0.1"}]
 // into [][]protocol.RvInstruction. It maps human-readable keys to RV variables
 // and converts protocol strings to the appropriate numeric code.
-func parseHumanReadableRvJSON(rawJSON []byte) ([][]protocol.RvInstruction, error) {
+// This is the V1 API format parser.
+// TODO: Remove this public export once #196 merges (V2 state layer migration complete)
```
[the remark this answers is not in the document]
sarmahaj: will be taken care in another PR (#196 )

internal/state/rvinfo_parser.go
sarmahaj: fixed, moved to utils.go now

internal/state/rvinfo_parser.go
sarmahaj: fixed, moved to utils.go now

internal/state/rvinfo.go
kgiusti: Sounds good to me - I'll resolve this conversation.

internal/state/rvinfo.go
kgiusti: Sounds good, I'll close out this request.

internal/state/rvinfo.go
kgiusti: Sounds good, I'll resolve this comment.

mmartinv (commented):
Let's try to refactor and simplify this a bit and try to do it as it is supposed to be done:

- Instead of checking on each request whether the database contains old V1 JSON or CBOR encoded struct let's handle the migration at database initialization time within the `InitDB` function within `internal/db/state.go` just after the `AutoMigrate` call.
- API handlers will be the responsible of converting their corresponding JSON to/from `[][]protocol.Instruction`. this means that all the `parse*` methods will go as private functions within each handler's package (`api/handlers` and `internal/handlers/rvinfo`)
- Persistence functions will just take `[][]protocol.Instruction` as an argument and save/load it to/from database using CBOR encoding, no JSON parsing.

does it make sense?

sarmahaj:
> Let's try to refactor and simplify this a bit and try to do it as it is supposed to be done:
> 
>     * Instead of checking on each request whether the database contains old V1 JSON or CBOR encoded struct let's handle the migration at database initialization time within the `InitDB` function within `internal/db/state.go` just after the `AutoMigrate` call.
> 
>     * API handlers will be the responsible of converting their corresponding JSON to/from `[][]protocol.Instruction`. this means that all the `parse*` methods will go as private functions within each handler's package (`api/handlers` and `internal/handlers/rvinfo`)
> 
>     * Persistence functions will just take `[][]protocol.Instruction` as an argument and save/load it to/from database using CBOR encoding, no JSON parsing.
> 
> 
> does it make sense?

@mmartinv This totally makes sense , its more clear and responsibilities are shared to each layer .
I will work on these and share the updated version.

internal/handlers/rvinfo/handler.go
sarmahaj: Fixed now ,ptal , thanks !

sarmahaj:
@mmartinv  @kgiusti 
Could you ptal , did some refactoring to separate out responsibilities of API layer parsing and persistence storage.
Remove the double parsing and other comments covered.

kgiusti:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

kgiusti (dismissed):
Thanks for addressing my review input, LGTM!

internal/handlers/rvinfo/handler.go
```diff
@@ -0,0 +1,162 @@
...
+			return UpdateRendezvousInfo500JSONResponse{
+				components.InternalServerError{Message: "failed to save rendezvous info"},
+			}, nil
+		}
+	} else if err != nil {
+		slog.Error("failed to update RV info", "error", err)
+		return UpdateRendezvousInfo500JSONResponse{
+			components.InternalServerError{Message: "failed to update rendezvous info"},
+		}, nil
+	}
```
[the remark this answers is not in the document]
sarmahaj: fixed as suggested

internal/handlers/rvinfo/parser.go
```diff
@@ -0,0 +1,367 @@
...
+			return nil, fmt.Errorf("delay_seconds must be a whole number, got %v", num)
+		}
+
+		// Validate non-negative
+		if num < 0 {
+			return nil, fmt.Errorf("delay_seconds must be non-negative, got %v", num)
+		}
+
+		secs := uint32(num)
+		enc, err := encodeRvValue(protocol.RVDelaysec, uint64(secs))
```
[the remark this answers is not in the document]
sarmahaj: fixed

mmartinv (approved):
Most of the feedback has been addressed, and the core functionality is now complete. While the code can benefit from some reorganization, I plan to handle that as part of the upcoming refactoring work.

LGTM. Thanks for working on this!

kgiusti (approved):

PR #214, opened 2026-04-24
fix: clean up upload directories and session state on TO2 failure

sarmahaj:
- Track GUID-based upload directories created during `fdo.upload` in TO2 and clean them up if the session fails or the device disconnects, preventing orphaned partial uploads from accumulating on disk.
  - Clean up related session state table rows (to2_sessions, to1_sessions, to0_sessions, etc.) in `InvalidateToken`, which previously left them orphaned on every failed session.

### DIFF
--- a/cmd/owner.go
+++ b/cmd/owner.go
@@ -340,10 +340,12 @@ type moduleStateMachines struct {
 }
 
 type moduleStateMachineState struct {
-	Name string
-	Impl serviceinfo.OwnerModule
-	Next func() (string, serviceinfo.OwnerModule, bool)
-	Stop func()
+	Name       string
+	Impl       serviceinfo.OwnerModule
+	Next       func() (string, serviceinfo.OwnerModule, bool)
+	Stop       func()
+	Completed  bool     // true when all modules have been yielded
+	UploadDirs []string // GUID-based upload directories created during this session
 }
 
 func (s moduleStateMachines) Module(ctx context.Context) (string, serviceinfo.OwnerModule, error) {
@@ -370,16 +372,18 @@ func (s moduleStateMachines) NextModule(ctx context.Context) (bool, error) {
 		if err != nil {
 			return false, fmt.Errorf("error getting devmod: %w", err)
 		}
-		next, stop := iter.Pull2(ownerModules(ctx, s.config, modules, s.DB))
-		module = &moduleStateMachineState{
-			Next: next,
-			Stop: stop,
-		}
+		module = &moduleStateMachineState{}
+		next, stop := iter.Pull2(ownerModules(ctx, s.config, modules, s.DB, &module.UploadDirs))
+		module.Next = next
+		module.Stop = stop
 		s.states[token] = module
 	}
 
 	var valid bool
 	module.Name, module.Impl, valid = module.Next()
+	if !valid {
+		module.Completed = true
+	}
 	return valid, nil
 }
 
@@ -393,10 +397,42 @@ func (s moduleStateMachines) CleanupModules(ctx context.Context) {
 		return
 	}
 	module.Stop()
+	s.cleanupUploadDirs(module)
 	delete(s.states, token)
+
+	// Sweep orphaned states from previous sessions that were invalidated
+	// by the transport layer (e.g. client disconnect) without a
+	// CleanupModules call.
+	s.sweepStaleStates()
 }
 
-func ownerModules(ctx context.Context, config *ServiceInfoConfig, modules []string, dbState *db.State) iter.Seq2[string, serviceinfo.OwnerModule] { //nolint:gocyclo
+// cleanupUploadDirs removes GUID-based upload directories if modules
+// did not complete naturally (i.e. TO2 failed).
+func (s moduleStateMachines) cleanupUploadDirs(module *moduleStateMachineState) {
+	if module.Completed {
+		return
+	}
+	for _, dir := range module.UploadDirs {
+		slog.Info("fdo.upload: cleaning up upload directory after TO2 failure", "dir", dir)
+		if err := os.RemoveAll(dir); err != nil {
+			slog.Error("fdo.upload: failed to clean up upload directory", "dir", dir, "err", err)
+		}
+	}
+}
+
+// sweepStaleStates iterates the states map and cleans up entries whose
+// sessions no longer exist in the database.
+func (s moduleStateMachines) sweepStaleStates() {
+	for token, module := range s.states {
+		if !s.DB.SessionExists(token) {
+			module.Stop()
+			s.cleanupUploadDirs(module)
+			delete(s.states, token)
+		}
+	}
+}
+
+func ownerModules(ctx context.Context, config *ServiceInfoConfig, modules []string, dbState *db.State, uploadDirs *[]string) iter.Seq2[string, serviceinfo.OwnerModule] { //nolint:gocyclo
 	return func(yield func(string, serviceinfo.OwnerModule) bool) {
 		if config == nil || len(config.Fsims) == 0 {
 			return
@@ -455,6 +491,11 @@ func ownerModules(ctx context.Context, config *ServiceInfoConfig, modules []stri
 						return
 					}
 					uploadDir = filepath.Join(uploadDir, hex.EncodeToString(replacementGUID[:]))
+
+					// Track the top-level GUID directory for cleanup on failure
+					if !slices.Contains(*uploadDirs, uploadDir) {
+						*uploadDirs = append(*uploadDirs, uploadDir)
+					}
 
 					// note: file.Dst has been validated as a relative path.
 					if file.Dst != "" {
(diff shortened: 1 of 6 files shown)


kgiusti:
Hi @sarmahaj - I'm playing with this fix and I've found a weird behavior. It may be due to a path not being covered by the cleanup path.  Still investigating, but I think you may be able to reproduce this.

I wanted to see what happens if TO2 fails due to a problem uploading a file.  So I modified the test/ci/test-fsim-upload.sh test to try to upload a file that doesn't exist. [Here](https://github.com/sarmahaj/go-fdo-server/blob/798efd474f807078a3e3a0ef2766cdbb67964399/test/ci/test-fsim-upload.sh#L46) I added a file to upload that does not exist:

```
diff --git a/test/ci/test-fsim-upload.sh b/test/ci/test-fsim-upload.sh
index 3b2a6cb..a05755c 100755
--- a/test/ci/test-fsim-upload.sh
+++ b/test/ci/test-fsim-upload.sh
@@ -44,6 +44,7 @@ owner:
             - src: "${device_files[1]}"
               dst: "${owner_files[1]}"
             - src: "${device_files[2]}"
+            - src: "nopeIdonotexist"
 EOF
 }
```

Then ran the test: `bash -c "source ./test/ci/test-fsim-upload.sh; run_test"`

I'm seeing TO2 fail:
```
[14:08:07] INFO: Attempting TO1 protocol
[14:08:08] INFO: TO1 succeeded
  base URL: http://rendezvous:8041
[14:08:08] INFO: Attempting TO2 protocol
[14:08:13] ERROR: TO2 failed
  base URL: http://owner:8043
  error: error reading KV to send to owner: could not read service info key: error uploading "nopeIdonotexist": open /home/kgiusti/work/fdo/go-fdo-server/test/workdir/device-credentials/nopeIdonotexist: no such file or directory
[14:08:18] ERROR: TO2 failed
  base URL: http://127.0.0.1:8043
  error: error reading KV to send to owner: could not read service info key: error uploading "nopeIdonotexist": open /home/kgiusti/work/fdo/go-fdo-server/test/workdir/device-credentials/nopeIdonotexist: no such file or directory
[14:08:18] INFO: Applying default delay for last directive
  delay: 1m55.013682632s
```
as expected.  This repeats... as expected.

What I found was that new GUID subdirectories are being created in the test upload directory `go-fdo-server/test/workdir/fsim/upload/owner/` each time the cycle repeats, but they are never cleaned up.  Each TO2 failure cycle adds a new GUID subdir, but never cleans up the previous one.

The code looks like it should be cleaning that up, but as best as I can tell the CleanupModule state machine callback is never being called by the go-fdo library.

I don't think this failure to call CleanupModule is related to this change at all.  Probably an issue elsewhere.  I'll poke some more... let you know what I see.

But related: maybe there's a good FSIM CI test we can add to test this.  Maybe an FSIM configuration that uploads a file, then runs an "fdo.command" that fails the first time it runs TO2, then passes the next time TO2 runs (saves some state to the filesystem to track this?).

kgiusti:
And here you go: https://github.com/fido-device-onboard/go-fdo/issues/231

That's preventing the owner from invoking the cleanup module path.  Here's a proposed fix:

https://github.com/fido-device-onboard/go-fdo/pull/232

sarmahaj:
@kgiusti Thanks for digging into this !
Your fix look good to me. https://github.com/fido-device-onboard/go-fdo/pull/232 

 In your test scenario, sweepStaleStates didn't help because it only runs inside CleanupModules, which was never called for any session due to the go-fdo bug. I am looking into adding an alternative trigger for the sweep (e.g., on new session creation) so it can catch orphans even when CleanupModules is never invoked.  (also until the fix lands on go-fdo)

Agree on adding a CI test for this/similar failure path.

sarmahaj:
### **Note**

The two new CI tests (test-upload-cleanup and test-upload-cleanup-missing-file) build both the client and server against  https://github.com/fido-device-onboard/go-fdo/pull/232 to validate the full cleanup flow end to end. Once that fix lands in go-fdo and we bump the dependency, these tests will build against the released version instead.

kgiusti (changes requested):
The tests you added look great - thanks!

One question and one _potential_ issue:

Q: Are the state.go changes related to the cleanup?  I'm having trouble understanding why those changes are part of this patch.

Issue: I haven't looked too closely at this code before but I think there's a potential race here over the moduleStateMachines struct.  This is a single-instance that is shared across all active TO2 sessions.  The http server spawns a separate goroutine for each client connection, which means they are all touching the same moduleStateMachines instance.

I think a lock needs to be added to this structure and taken when the `states` map and `lastSweep` flags are being referenced.

kgiusti:
Ah - and if I paid attention to the gemini review comments I would've seen that it already raised the race issue... ouch.

cmd/owner.go
```diff
@@ -393,10 +407,37 @@ func (s moduleStateMachines) CleanupModules(ctx context.Context) {
...
+// cleanupUploadDirs removes GUID-based upload directories if modules
+// did not complete naturally (i.e. TO2 failed).
+func (s *moduleStateMachines) cleanupUploadDirs(module *moduleStateMachineState) {
+	if module.Completed {
+		return
+	}
+	for _, dir := range module.UploadDirs {
+		slog.Info("fdo.upload: cleaning up upload directory after TO2 failure", "dir", dir)
+		if err := os.RemoveAll(dir); err != nil {
+			slog.Error("fdo.upload: failed to clean up upload directory", "dir", dir, "err", err)
```
djach7, line 435: Would it make sense to leave a comment here that this is a best effort cleanup? Given that if the cleanup fails and this error is passed the directory still remains.

I suppose the other way to do it would be to track it for retry or something like that, but I think a comment as a disclaimer suffices.

cmd/owner.go
```diff
@@ -358,32 +363,41 @@ func (s moduleStateMachines) Module(ctx context.Context) (string, serviceinfo.Ow
...
+func (s *moduleStateMachines) NextModule(ctx context.Context) (bool, error) {
 	token, ok := s.DB.TokenFromContext(ctx)
 	if !ok {
 		return false, fmt.Errorf("invalid context: no token")
 	}
 	module, ok := s.states[token]
 	if !ok {
+		// Sweep orphaned states from previous sessions before creating a
+		// new one. Throttled to avoid excessive DB queries under load.
+		if time.Since(s.lastSweep) >= sweepInterval {
```
djach7, line 383: I think this might result in a similar race issue to the one gemini documented elsewhere. The mutex fix gemini suggested for the states race should be able to fix this one too.

djach7 (changes requested):
Two relatively small changes, nothing major

sarmahaj:
> The tests you added look great - thanks!
> 
> One question and one _potential_ issue:
> 
> Q: Are the state.go changes related to the cleanup? I'm having trouble understanding why those changes are part of this patch.

@kgiusti 
The state.go changes handle the database side of the same orphan problem. Previously InvalidateToken only deleted the Session row, leaving child rows (TO2Session, KeyExchange, etc.) orphaned in every failed or timed-out session. The upload dir cleanup handles filesystem orphans, and the InvalidateToken change handles database orphans. Both are triggered by the same TO2 failure/disconnect path, so they made sense together. Happy to split into a separate commit if you'd prefer clearer separation.                                                                                                    

> 
> Issue: I haven't looked too closely at this code before but I think there's a potential race here over the moduleStateMachines struct. This is a single-instance that is shared across all active TO2 sessions. The http server spawns a separate goroutine for each client connection, which means they are all touching the same moduleStateMachines instance.
> 
> I think a lock needs to be added to this structure and taken when the `states` map and `lastSweep` flags are being referenced.

you're right. The states map was already shared across goroutines, but this PR adds lastSweep and sweepStaleStates() which iterates the map and deletes stale entries, increasing the exposure. I'll add a sync.Mutex to the struct and lock in Module(), NextModule(), and CleanupModules() to address this.

cmd/owner.go
sarmahaj: thanks for pointing that ,added a mutex lock now

cmd/owner.go
sarmahaj: Good point — updated the doc comment to clarify this is best-effort cleanup. If os.RemoveAll fails, the directory is left in place and the error is logged. There's no retry path since the entry is removed from the in-memory states map regardless of outcome, so we lose track of it. A startup sweep for orphaned GUID subdirs could address that, but seems better as a follow-up.

cmd/owner.go
```diff
@@ -387,10 +391,42 @@ func (s moduleStateMachines) CleanupModules(ctx context.Context) {
 		return
 	}
 	module.Stop()
+	s.cleanupUploadDirs(module)
 	delete(s.states, token)
+
+	// Sweep orphaned states from previous sessions that were invalidated
+	// by the transport layer (e.g. client disconnect) without a
+	// CleanupModules call.
+	s.sweepStaleStates()
```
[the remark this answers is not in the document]
sarmahaj: fixed in commit 0ecf5a3. The sweep was moved from CleanupModules to NextModule and throttled with sweepInterval (1 minute)

PR #229, opened 2026-05-08
ci: implement least-privileged token permissions on workflows

kgiusti:
Explicitly set the workflow permissions to read-only at the topmost level. Override where necessary at the run level.

Closes: #227

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -6,6 +6,9 @@ on:
   schedule:
     - cron: '19 20 * * 6'
 
+permissions:
+  contents: read
+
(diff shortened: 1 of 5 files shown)


knecasov (commented):
LGTM. I have just one note, I would replace "priviledged" with "privileged" in the commit message and the PR title.

kgiusti:
> LGTM. I have just one note, I would replace "priviledged" with "privileged" in the commit message and the PR title.

English isn't my primary language.
It's "C"  :)

knecasov (approved):
LGTM, thank you very much for fixing!

PR #239, opened 2026-06-03
ci: support testing from brew builds and compose repositories

mcattamoredhat:
## Summary

This PR extends the test infrastructure to allow installing RPM packages from two new sources — internal brew builds and compose repositories — for `test/rpm` and `test/bootc` deployments. It also includes three follow-up fixes for bugs found during review and manual/CI testing of that new support.

---

## Commit 1 — `test: allow testing brew builds`

Introduces support for installing pre-built RPMs from an internal brew server by pointing at a specific build's version/release directory.

**Usage:**

```
SERVER_RPM_URL="http://brew.host/packages/go-fdo-server/1.0.0/1.el10/x86_64" \
CLIENT_RPM_URL="http://brew.host/packages/go-fdo-client/1.0.0/1.el10/x86_64" \
  sudo ./test/rpm/test-onboarding.sh
```

---

## Commit 2 — `ci: add compose repository support for RPM and bootc tests`

Introduces support for installing RPM packages from a compose repository URL, targeting a specific compose snapshot.

**Usage:**

```
# RPM deployment
sudo COMPOSE_BASE_URL="http://<host>/.../RHEL-10.2-<date>/compose/" \
     ./test/rpm/test-onboarding.sh
# bootc deployment
sudo COMPOSE_BASE_URL="http://<host>/.../RHEL-10.2-<date>/compose/" \
     ./test/bootc/test-onboarding.sh
# Override streams if needed (default: BaseOS AppStream for RHEL/CentOS)
sudo COMPOSE_BASE_URL="..." COMPOSE_STREAMS="BaseOS AppStream CRB" \
     ./test/rpm/test-onboarding.sh
```

---

## Commit 3 — `fix: avoid unbound _brew_* vars in parse_brew_url callers`

`parse_brew_url()` set `_brew_ver`/`_brew_rel`/`_brew_arch` as side effects, but every caller captured it via `url=$(parse_brew_url ...)`. Command substitution runs the function in a subshell, so those assignments never reached the calling scope — under `set -u`, the first reference to `_brew_arch` in `install_client()`/`install_server()` (`test/ci`, `test/rpm`, `test/bootc`) crashed with `"unbound variable"` whenever `SERVER_RPM_URL` or `CLIENT_RPM_URL` was actually set.

**Fix:** have `parse_brew_url()` set `_brew_url` directly instead of echoing it, and call it as a plain statement (no command substitution) so all four variables land in the caller's scope.

No behavior change for any path that doesn't set `SERVER_RPM_URL`/`CLIENT_RPM_URL`: `PACKIT_COPR_RPMS`-driven Packit/Testing Farm runs and the GitHub Actions `test/ci` matrix never touch `parse_brew_url()` at all — this only affects the manual brew-build testing path introduced in Commit 1.

---

## Commit 4 — `fix: strip trailing slash from COMPOSE_BASE_URL and drop dead repo file`

- `install_from_compose()` (`test/rpm/utils.sh`) and the bootc client compose branch (`test/bootc/utils.sh`) now strip a trailing slash from the compose base URL before building `baseurl=...`, avoiding a double slash if the caller passes `COMPOSE_BASE_URL` with one.
- Removed `files/rhel-10-2.repo`: an unused static template with unreplaced `REPLACE_ME_HERE` placeholders. The real RHEL repo file is generated at runtime as `files/rhel-${VERSION_ID}.repo` (`test/bootc/utils.sh`), so this static file was dead weight.

---

## Commit 5 — `fix(bootc): wait for firewalld D-Bus before starting libvirt network`

`configure_service_firewalld()` returned right after `systemctl start firewalld`, but firewalld ≥ 2.4.1 (e.g. on Fedora 44) no longer blocks until its D-Bus interface is ready. `configure_service_libvirtd()` runs immediately afterward and calls `virsh net-start integration` on a network with `zone='trusted'`, which requires firewalld's D-Bus API — racing it produces:

```
error: internal error: zone trusted requested for network integration but firewalld is not active
```

**Fix:** poll `firewall-cmd --state` (up to 30s) after starting firewalld, bailing out with diagnostics if the unit dies or the timeout is hit. Same root cause/fix as [`virt-s1/rhel-edge#12261`](https://github.com/virt-s1/rhel-edge/pull/12261/changes).


mcattamoredhat:
Opened as draft PR for testing purposes only.
This PR will be opened again in the proper repository `go-fdo-ci`, after making sure it is useful for local testing.

mcattamoredhat:
Closing this PR and opening https://github.com/fido-device-onboard/go-fdo-ci/pull/20

PR #243, opened 2026-06-30
docs: Management API Auth Phase 1 — design spec and implementation plan

mmartinv:
## Summary

- Design specification for Phase 1 of the management API authentication and authorization system
- Implementation plan with 10 detailed tasks covering the full Phase 1 scope

## Context

Addresses the **THEEDGE-4735** spike: defines roles/personas, permissions, and authentication mechanisms for the FDO server management API.

### Phase 1 scope
- Auth framework: pluggable `Authenticator` interface, `Identity` model, AuthN/AuthZ middleware chain
- API key authentication mechanism (first mechanism, validates full pipeline)
- RBAC + fine-grained scopes (`resource:read`/`write`/`delete` pattern)
- Database models: users, roles, role scopes, API keys (GORM, same DB as server data)
- Configuration schema (`auth` block in YAML, multiple mechanisms simultaneously)
- OpenAPI `x-required-scopes` custom extension as single source of truth for route authorization
- Bootstrapping: config-file seed + CLI `init-admin` command
- Backward compatible: auth disabled by default

### Key design decisions updated from review feedback

**Cryptography & authentication:**
- **SHA-256 for API key hashing** (not bcrypt) — API keys have ~190 bits of entropy, making bcrypt's intentional slowness unnecessary and a DoS risk. Uses `crypto/sha256` + `crypto/subtle.ConstantTimeCompare`
- **User.Active check during authentication** — deactivating a user immediately revokes all their API keys
- **Uniform error responses** — expired keys return the same error as invalid keys to avoid leaking key validity to attackers

**Path security:**
- **Exact-match path exclusion** — `/health` excludes only `/health`, not `/health-check`. Trailing `/` enables prefix matching (e.g., `/api/docs/`)
- **Path normalization** — `path.Clean()` applied before exclusion checks to prevent bypass via `//health` or dot segments
- **V1 API auth protection** — legacy `/api/v1/` endpoints wrapped with `authN` middleware (explicit code, not just a note) when auth is enabled

**Context & lifecycle:**
- **Full context.Context propagation** — all 12 state functions accept `ctx` and use `db.WithContext(ctx)` for request cancellation and tracing. Background goroutines use `context.WithoutCancel` to preserve trace IDs
- **Graceful shutdown** — `APIKeyAuthenticator` uses `sync.WaitGroup` to drain background goroutines; 5-second timeout on async `UpdateAPIKeyLastUsed`

**Robustness:**
- **Atomic seed operation** — `SeedAuth` wraps all operations in a single transaction; `FindOrCreateRole` handles re-seeding with `--force`; preset key validated for `fdo_` prefix
- **No panics** — `Handler()` returns `(http.Handler, error)` instead of panicking on scope parse failures
- **Consistent error wrapping** — all state functions wrap errors with `fmt.Errorf("...: %w", err)` for debuggable stack traces
- **DB errors logged in authenticator** — `FindAPIKeysByPrefix`, `GetUserByID`, `GetUserRoles` failures logged at ERROR level (not silently swallowed)
- **Safe slice operations** — `defaultExcludedPaths` uses `make`+`copy` to prevent aliasing; `SeedAuth` uses full slice expression for scope append

**Code quality:**
- **Compile-time interface check** — `var _ auth.Authenticator = (*APIKeyAuthenticator)(nil)`
- **Per-test SQLite isolation** — each test uses a unique named in-memory database via `t.Name()`, preventing cross-test pollution
- **Seed API key output** — printed to stderr via `fmt.Fprintf`, not through `slog` (avoids leaking keys to log aggregation systems)

### Future phases
- **Phase 2**: Management API endpoints (CRUD for users/groups/roles/API keys)
- **Phase 3**: Additional auth mechanisms (Bearer, Basic, OIDC, OAuth2, SAML)

## Files
- `docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md` — design specification
- `docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md` — implementation plan

## Test plan
- [x] Review design spec for completeness and correctness
- [x] Review implementation plan tasks for feasibility
- [x] Verify scope definitions cover all existing V2 API endpoints
- [x] Verify backward compatibility guarantees are sound
- [x] Address Gemini Code Assist security findings (5 items)
- [x] Address reviewer feedback (11 items from knecasov)
- [x] Go design patterns review (5 passes)
- [x] Go safety review (slice aliasing, bounds checks, test correctness)
- [x] Go context propagation review (all DB calls context-aware)
- [x] Go security audit (DREAD-scored, 4 findings addressed)
- [x] Go error handling review (consistent wrapping, no swallowed errors)
- [x] Go testing review (inactive user, path exclusion, per-test isolation)
- [x] Final 6-agent parallel verification (no remaining issues)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
+++ b/docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
@@ -1,0 +1,3554 @@
+# Management API Auth Phase 1 Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add authentication (API key) and authorization (RBAC + scopes) to the FDO server management API, with a pluggable framework for future auth mechanisms.
+
+**Architecture:** Middleware chain (AuthN → AuthZ) inserted between rate limiter and OpenAPI validation. Scopes parsed from `x-required-scopes` OpenAPI extensions at startup. API key authenticator as the first pluggable mechanism. Auth disabled by default for backward compatibility.
+
+**Tech Stack:** Go 1.26, GORM (SQLite/PostgreSQL), `crypto/sha256` + `crypto/subtle` (stdlib), `github.com/getkin/kin-openapi`, Cobra/Viper, oapi-codegen
+
+## Global Constraints
+
+- Module path: `github.com/fido-device-onboard/go-fdo-server`
+- Existing GORM model pattern: models in `internal/state/`, init functions return `(*State, error)`, `AutoMigrate` called in init
+- Existing middleware pattern: `func(http.Handler) http.Handler` closures in `internal/middleware/`
+- Existing config pattern: structs with `mapstructure` tags in `internal/config/`, validated via `.Validate()` methods
+- Existing CLI pattern: Cobra commands in `cmd/`, Viper for config binding
+- OpenAPI specs: per-resource YAML in `api/v2/<resource>/openapi.yaml`, merged into `openapi.json` via `go:generate npx openapi-format`
+- FDO protocol endpoints (`/fdo/101/msg/*`, `/fdo/200/msg/*`), `/health`, `/api/docs/*`, `/api/openapi.json` are NEVER subject to management API auth
+- Default behavior when `auth` is not configured: all endpoints open (backward compatible)
+- SPDX header: `// SPDX-FileCopyrightText: (C) 2026 Red Hat Inc.` + `// SPDX-License-Identifier: Apache 2.0`
+- Error responses: `{"message": "..."}` JSON format, matching existing `errorResponse` struct pattern
+
+---
+
+### Task 1: Database Models for Users, Roles, and API Keys
+
+**Files:**
+- Create: `internal/state/user.go`
+- Create: `internal/state/role.go`
+- Create: `internal/state/apikey.go`
+- Create: `internal/state/auth.go`
+- Create: `internal/state/user_test.go`
+- Create: `internal/state/role_test.go`
+- Create: `internal/state/auth_test.go`
+- Create: `internal/state/apikey_test.go`
+
+**Interfaces:**
+- Consumes: `gorm.io/gorm` (existing dependency)
+- Produces:
+  - `type User struct` — GORM model with ID (UUID string), Name, Email, Active, CreatedAt, UpdatedAt
+  - `type Role struct` — GORM model with ID (UUID string), Name, Description, BuiltIn (bool), CreatedAt, UpdatedAt
+  - `type UserRole struct` — join table with UserID, RoleID
+  - `type RoleScope struct` — GORM model with RoleID, Scope (string)
+  - `type APIKey struct` — GORM model with ID (UUID string), Prefix (6 chars), HashedKey ([]byte), Name, UserID (FK), Scopes (JSON string slice), ExpiresAt (*time.Time), Active (bool), LastUsedAt (*time.Time), CreatedAt, UpdatedAt
+  - `func InitAuthDB(ctx context.Context, db *gorm.DB) error` — runs AutoMigrate for all auth models
+  - `func CreateUser(ctx context.Context, db *gorm.DB, name, email string) (*User, error)`
+  - `func GetUserByID(ctx context.Context, db *gorm.DB, id string) (*User, error)`
+  - `func GetUserRoles(ctx context.Context, db *gorm.DB, userID string) ([]Role, error)`
+  - `func GetUserScopes(ctx context.Context, db *gorm.DB, userID string) ([]string, error)`
+  - `func CreateRole(ctx context.Context, db *gorm.DB, name, description string, builtIn bool, scopes []string) (*Role, error)`
+  - `func AssignRoleToUser(ctx context.Context, db *gorm.DB, userID, roleID string) error`
+  - `func CreateAPIKey(ctx context.Context, db *gorm.DB, name, userID string, scopes []string, expiresAt *time.Time) (*APIKey, string, error)` — returns model + cleartext key
+  - `func FindAPIKeysByPrefix(ctx context.Context, db *gorm.DB, prefix string) ([]APIKey, error)`
+  - `func UpdateAPIKeyLastUsed(ctx context.Context, db *gorm.DB, id string)`
+  - `func UserCount(ctx context.Context, db *gorm.DB) (int64, error)`
+
+- [ ] **Step 1: Write test for User CRUD**
+
+Create `internal/state/user_test.go`:
+
+```go
+// SPDX-FileCopyrightText: (C) 2026 Red Hat Inc.
+// SPDX-License-Identifier: Apache 2.0
+
+package state
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+func setupAuthTestDB(t *testing.T) *gorm.DB {
+	t.Helper()
+	dsn := fmt.Sprintf("file:%s?mode=memory&cache=shared", strings.ReplaceAll(t.Name(), "/", "_"))
+	db, err := gorm.Open(sqlite.Open(dsn), &gorm.Config{})
+	if err != nil {
+		t.Fatalf("Failed to open test database: %v", err)
+	}
+	if err := db.Exec("PRAGMA foreign_keys = ON").Error; err != nil {
+		t.Fatalf("Failed to enable foreign keys: %v", err)
+	}
+	if err := InitAuthDB(t.Context(), db); err != nil {
+		t.Fatalf("Failed to initialize auth database: %v", err)
+	}
+	return db
+}
+
+func TestCreateUser(t *testing.T) {
+	db := setupAuthTestDB(t)
+	ctx := t.Context()
+	user, err := CreateUser(ctx, db, "admin", "admin@example.com")
+	if err != nil {
+		t.Fatalf("CreateUser failed: %v", err)
+	}
+	if user.ID == "" {
+		t.Error("expected non-empty ID")
+	}
+	if user.Name != "admin" {
+		t.Errorf("expected name 'admin', got %q", user.Name)
+	}
+	if user.Email != "admin@example.com" {
+		t.Errorf("expected email 'admin@example.com', got %q", user.Email)
+	}
+	if !user.Active {
+		t.Error("expected user to be active")
+	}
+}
+
+func TestGetUserByID(t *testing.T) {
+	db := setupAuthTestDB(t)
+	ctx := t.Context()
+	created, err := CreateUser(ctx, db, "admin", "admin@example.com")
+	if err != nil {
+		t.Fatalf("CreateUser failed: %v", err)
+	}
+
+	found, err := GetUserByID(ctx, db, created.ID)
+	if err != nil {
+		t.Fatalf("GetUserByID failed: %v", err)
+	}
+	if found.ID != created.ID {
+		t.Errorf("expected ID %q, got %q", created.ID, found.ID)
+	}
+	if found.Name != "admin" {
+		t.Errorf("expected name 'admin', got %q", found.Name)
+	}
+}
+
+func TestGetUserByID_NotFound(t *testing.T) {
+	db := setupAuthTestDB(t)
+	_, err := GetUserByID(t.Context(), db, "nonexistent")
+	if err == nil {
+		t.Error("expected error for nonexistent user")
+	}
+}
+
+func TestUserCount(t *testing.T) {
+	db := setupAuthTestDB(t)
+	ctx := t.Context()
+	count, err := UserCount(ctx, db)
+	if err != nil {
+		t.Fatalf("UserCount failed: %v", err)
+	}
+	if count != 0 {
+		t.Errorf("expected count 0, got %d", count)
+	}
+
+	if _, err := CreateUser(ctx, db, "admin", "admin@example.com"); err != nil {
+		t.Fatalf("CreateUser failed: %v", err)
+	}
+
+	count, err = UserCount(ctx, db)
+	if err != nil {
+		t.Fatalf("UserCount failed: %v", err)
+	}
+	if count != 1 {
+		t.Errorf("expected count 1, got %d", count)
+	}
+}
+
+func TestCreateUser_Validation(t *testing.T) {
+	db := setupAuthTestDB(t)
+	ctx := t.Context()
+
+	tests := []struct {
+		name  string
+		uname string
+		email string
+	}{
+		{"empty name", "", "test@example.com"},
+		{"empty email", "admin", ""},
+		{"invalid email", "admin", "not-an-email"},
+		{"display name form", "admin", "Admin <admin@example.com>"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := CreateUser(ctx, db, tt.uname, tt.email)
+			if err == nil {
+				t.Error("expected validation error")
+			}
+		})
+	}
+}
+```
+
+- [ ] **Step 2: Run test to verify it fails**
+
+Run: `go test ./internal/state/ -run "TestCreateUser|TestGetUserByID|TestUserCount" -v`
+Expected: FAIL — `InitAuthDB`, `CreateUser`, `GetUserByID`, `UserCount` not defined
+
+- [ ] **Step 3: Implement User model and operations**
+
+Create `internal/state/user.go`:
+
+```go
+// SPDX-FileCopyrightText: (C) 2026 Red Hat Inc.
+// SPDX-License-Identifier: Apache 2.0
+
+package state
+
+import (
+	"context"
+	"fmt"
+	"net/mail"
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+	"gorm.io/gorm"
+)
+
+type User struct {
+	ID        string    `gorm:"type:varchar(36);primaryKey"`
+	Name      string    `gorm:"type:varchar(255);not null"`
+	Email     string    `gorm:"type:varchar(255);not null;uniqueIndex"`
+	Active    bool      `gorm:"type:boolean;not null"`
+	CreatedAt time.Time `gorm:"autoCreateTime:milli"`
+	UpdatedAt time.Time `gorm:"autoUpdateTime:milli"`
+}
+
+func (User) TableName() string { return "users" }
+
+func CreateUser(ctx context.Context, db *gorm.DB, name, email string) (*User, error) {
+	name = strings.TrimSpace(name)
+	email = strings.TrimSpace(email)
+	if name == "" {
+		return nil, fmt.Errorf("user name must not be empty")
+	}
+	if len(name) > 255 {
+		return nil, fmt.Errorf("user name must not exceed 255 characters")
+	}
+	if len(email) > 255 {
+		return nil, fmt.Errorf("email must not exceed 255 characters")
+	}
+	addr, err := mail.ParseAddress(email)
+	if err != nil || addr.Address != email {
+		return nil, fmt.Errorf("invalid email address: %q", email)
+	}
+
+	user := &User{
+		ID:     uuid.New().String(),
+		Name:   name,
+		Email:  email,
+		Active: true,
+	}
+	if err := db.WithContext(ctx).Create(user).Error; err != nil {
+		return nil, fmt.Errorf("failed to create user: %w", err)
+	}
+	return user, nil
+}
+
+func GetUserByID(ctx context.Context, db *gorm.DB, id string) (*User, error) {
+	var user User
+	if err := db.WithContext(ctx).Where("id = ?", id).First(&user).Error; err != nil {
+		return nil, fmt.Errorf("failed to get user: %w", err)
+	}
+	return &user, nil
+}
+
+func UserCount(ctx context.Context, db *gorm.DB) (int64, error) {
+	var count int64
+	if err := db.WithContext(ctx).Model(&User{}).Count(&count).Error; err != nil {
+		return 0, fmt.Errorf("failed to count users: %w", err)
+	}
+	return count, nil
+}
+```
+
+- [ ] **Step 4: Run user tests to verify they pass**
+
+Run: `go test ./internal/state/ -run "TestCreateUser|TestGetUserByID|TestUserCount" -v`
+Expected: PASS
+
+- [ ] **Step 5: Write test for Role CRUD**
+
+Create `internal/state/role_test.go`:
+
+```go
+// SPDX-FileCopyrightText: (C) 2026 Red Hat Inc.
+// SPDX-License-Identifier: Apache 2.0
+
+package state
+
+import (
+	"slices"
+	"testing"
+)
+
+func TestCreateRole(t *testing.T) {
+	db := setupAuthTestDB(t)
+	ctx := t.Context()
+	role, err := CreateRole(ctx, db, "admin", "Full access", true, []string{"vouchers:read", "vouchers:write", "vouchers:delete"})
+	if err != nil {
+		t.Fatalf("CreateRole failed: %v", err)
+	}
+	if role.ID == "" {
+		t.Error("expected non-empty ID")
+	}
+	if role.Name != "admin" {
+		t.Errorf("expected name 'admin', got %q", role.Name)
+	}
+	if !role.BuiltIn {
+		t.Error("expected BuiltIn to be true")
+	}
+}
+
+func TestAssignRoleAndGetScopes(t *testing.T) {
+	db := setupAuthTestDB(t)
+	ctx := t.Context()
+	expectedScopes := []string{"vouchers:read", "vouchers:write", "vouchers:delete"}
+	role, err := CreateRole(ctx, db, "admin", "Full access", true, expectedScopes)
+	if err != nil {
+		t.Fatalf("CreateRole failed: %v", err)
+	}
+
+	user, err := CreateUser(ctx, db, "testuser", "test@example.com")
+	if err != nil {
+		t.Fatalf("CreateUser failed: %v", err)
+	}
+
+	if err := AssignRoleToUser(ctx, db, user.ID, role.ID); err != nil {
+		t.Fatalf("AssignRoleToUser failed: %v", err)
+	}
+
+	roles, err := GetUserRoles(ctx, db, user.ID)
+	if err != nil {
+		t.Fatalf("GetUserRoles failed: %v", err)
+	}
+	if len(roles) != 1 {
+		t.Fatalf("expected 1 role, got %d", len(roles))
+	}
+	if roles[0].Name != "admin" {
+		t.Errorf("expected role name 'admin', got %q", roles[0].Name)
+	}
+
+	scopes, err := GetUserScopes(ctx, db, user.ID)
+	if err != nil {
+		t.Fatalf("GetUserScopes failed: %v", err)
+	}
+	slices.Sort(scopes)
+	slices.Sort(expectedScopes)
+	if !slices.Equal(scopes, expectedScopes) {
+		t.Errorf("expected scopes %v, got %v", expectedScopes, scopes)
+	}
+}
+
+func TestFindOrCreateRole(t *testing.T) {
+	db := setupAuthTestDB(t)
+	ctx := t.Context()
+
+	role1, err := FindOrCreateRole(ctx, db, "admin", "Full access", true, []string{"vouchers:read"})
+	if err != nil {
+		t.Fatalf("FindOrCreateRole (create) failed: %v", err)
+	}
+
+	role2, err := FindOrCreateRole(ctx, db, "admin", "Full access", true, []string{"vouchers:read"})
+	if err != nil {
+		t.Fatalf("FindOrCreateRole (find) failed: %v", err)
+	}
+
+	if role1.ID != role2.ID {
+		t.Errorf("expected same role ID, got %q and %q", role1.ID, role2.ID)
+	}
+}
+
+func TestCreateRole_Validation(t *testing.T) {
+	db := setupAuthTestDB(t)
+	ctx := t.Context()
+
+	_, err := CreateRole(ctx, db, "", "description", false, nil)
+	if err == nil {
+		t.Error("expected validation error for empty role name")
+	}
+
+	_, err = CreateRole(ctx, db, "admin", "description", false, []string{"INVALID:SCOPE"})
+	if err == nil {
+		t.Error("expected validation error for invalid scope format")
+	}
+}
+```
+
+- [ ] **Step 6: Implement Role model and operations**
+
+Create `internal/state/role.go`:
+
+```go
+// SPDX-FileCopyrightText: (C) 2026 Red Hat Inc.
+// SPDX-License-Identifier: Apache 2.0
+
+package state
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+	"gorm.io/gorm"
+)
+
+type Role struct {
+	ID          string    `gorm:"type:varchar(36);primaryKey"`
+	Name        string    `gorm:"type:varchar(255);not null;uniqueIndex"`
+	Description string    `gorm:"type:text"`
+	BuiltIn     bool      `gorm:"type:boolean;not null;default:false"`
+	CreatedAt   time.Time `gorm:"autoCreateTime:milli"`
+	UpdatedAt   time.Time `gorm:"autoUpdateTime:milli"`
+}
+
+func (Role) TableName() string { return "roles" }
+
+type RoleScope struct {
+	RoleID  string `gorm:"type:varchar(36);not null;primaryKey"`
+	Scope   string `gorm:"type:varchar(255);not null;primaryKey"`
+	RoleRef *Role  `gorm:"foreignKey:RoleID;references:ID;constraint:OnDelete:CASCADE"`
+}
+
+func (RoleScope) TableName() string { return "role_scopes" }
+
+type UserRole struct {
+	UserID  string `gorm:"type:varchar(36);not null;primaryKey"`
+	RoleID  string `gorm:"type:varchar(36);not null;primaryKey"`
+	UserRef *User  `gorm:"foreignKey:UserID;references:ID;constraint:OnDelete:CASCADE"`
+	RoleRef *Role  `gorm:"foreignKey:RoleID;references:ID;constraint:OnDelete:CASCADE"`
+}
+
+func (UserRole) TableName() string { return "user_roles" }
+
+func CreateRole(ctx context.Context, db *gorm.DB, name, description string, builtIn bool, scopes []string) (*Role, error) {
+	name = strings.TrimSpace(name)
+	if name == "" {
+		return nil, fmt.Errorf("role name must not be empty")
+	}
+	if len(name) > 255 {
+		return nil, fmt.Errorf("role name must not exceed 255 characters")
+	}
+	if len(description) > 1024 {
+		return nil, fmt.Errorf("role description must not exceed 1024 characters")
+	}
+	for _, s := range scopes {
+		if !validScopePattern.MatchString(s) {
+			return nil, fmt.Errorf("invalid scope %q: must match [a-z0-9][a-z0-9:_-]*", s)
+		}
+	}
+
+	role := &Role{
+		ID:          uuid.New().String(),
+		Name:        name,
+		Description: description,
+		BuiltIn:     builtIn,
+	}
+	uniqueScopes := make([]string, 0, len(scopes))
+	seen := make(map[string]struct{})
+	for _, s := range scopes {
+		if _, ok := seen[s]; !ok {
+			seen[s] = struct{}{}
+			uniqueScopes = append(uniqueScopes, s)
+		}
+	}
+	err := db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		if err := tx.Create(role).Error; err != nil {
+			return fmt.Errorf("failed to create role: %w", err)
+		}
+		if len(uniqueScopes) > 0 {
+			roleScopes := make([]RoleScope, len(uniqueScopes))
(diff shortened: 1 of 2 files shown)


docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,518 @@
...
+2. **Unknown scopes error**: `x-required-scopes` references a scope not in any role definition → fail startup
+3. **Unprotected endpoints info**: Log which endpoints have no `x-required-scopes`
+
+### Files Requiring `x-required-scopes` Annotations
+
+| File | Operations |
+|---|---|
+| `api/v2/voucher/openapi.yaml` | 6 operations |
+| `api/v2/deviceca/openapi.yaml` | 4 operations |
+| `api/v2/rvinfo/openapi.yaml` | 2 operations |
```
knecasov: It seems there should be 3 operations. The scope table on [lines 104-106](https://github.com/fido-device-onboard/go-fdo-server/pull/243/changes#diff-15b683bed55dfd87553084f16b242e7ef83691ec1ed2bf12529603882e42d443R104-R106) defines 3 `rvinfo` scopes (`read`, `write`, `delete`).

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,518 @@
...
+```
+
+An API key can never exceed the user's permissions, even if the user's roles change after key creation.
+
+### Authentication Flow
+
+1. Extract `X-API-Key` header — if absent, return `nil, nil` (try next authenticator)
+2. Validate format (`fdo_` prefix, minimum length) — if invalid, return `nil, error`
+3. Extract prefix (6 chars after `fdo_`)
+4. Query candidates by prefix where `active = true`
```
knecasov: Does `active = true` refer to `APIKey.Active`, `User.Active` ([line 150](https://github.com/fido-device-onboard/go-fdo-server/pull/243/changes#diff-15b683bed55dfd87553084f16b242e7ef83691ec1ed2bf12529603882e42d443R150)), or both, please? If it is only the API key, deactivated user's keys would uselessly continue in the authentication flow. Consider clarifying.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,518 @@
...
+API key scopes (if set) → intersect with user's full scope set = effective scopes
+API key scopes (if empty) → user's full scope set = effective scopes
+```
+
+An API key can never exceed the user's permissions, even if the user's roles change after key creation.
+
+### Authentication Flow
+
+1. Extract `X-API-Key` header — if absent, return `nil, nil` (try next authenticator)
+2. Validate format (`fdo_` prefix, minimum length) — if invalid, return `nil, error`
```
knecasov: I would specify what the minimum length means.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,518 @@
...
+    // Name returns the mechanism name (e.g., "api-key", "oidc")
+    Name() string
+    // Authenticate inspects the request and returns an Identity if credentials are valid.
+    // Returns nil, nil if this authenticator doesn't match the request (no relevant credentials present).
+    // Returns nil, error if credentials are present but invalid.
+    Authenticate(ctx context.Context, r *http.Request) (*Identity, error)
+}
+```
+
+The semantic: `nil, nil` means "not my request, try the next authenticator." `nil, error` means "credentials were present but invalid — stop and return 401."
```
knecasov: Dots at the end of strings seem to be unintentional. What about the following format?

The semantic: `nil, nil` means "not my request, try the next authenticator"; `nil, error` means "credentials were present but invalid — stop and return 401".

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,518 @@
...
+**Manufacturer server:**
+
+| Scope | Description |
+|---|---|
+| `vouchers:read` | List/get vouchers |
+| `vouchers:write` | Import vouchers |
+| `vouchers:delete` | Delete vouchers |
+| `vouchers:extend` | Extend voucher ownership |
+| `rvinfo:read` | Read rendezvous info config |
+| `rvinfo:write` | Update rendezvous info config |
```
knecasov: Should not be also "Create" in the Description column?

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,518 @@
...
+| `rvinfo:delete` | Delete rendezvous info config |
+
+**Owner server:**
+
+| Scope | Description |
+|---|---|
+| `vouchers:read` | List/get vouchers |
+| `vouchers:write` | Import/verify-ownership vouchers |
```
knecasov: I would write "verify ownership" instead of "verify-ownership".

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,518 @@
...
+### Files Requiring `x-required-scopes` Annotations
+
+| File | Operations |
+|---|---|
+| `api/v2/voucher/openapi.yaml` | 6 operations |
+| `api/v2/deviceca/openapi.yaml` | 4 operations |
+| `api/v2/rvinfo/openapi.yaml` | 2 operations |
+| `api/v2/rvto2addr/openapi.yaml` | 3 operations |
+| `api/v2/device/openapi.yaml` | 1 operation |
+| `api/v2/health/openapi.yaml` | 0 (public) |
```
knecasov: I would write "0 operations (public)".

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,518 @@
...
+Added to each server's YAML config under a new `auth` key:
+
+```yaml
+auth:
+  # When false or omitted, all endpoints are accessible without authentication.
+  # When true, at least one mechanism must be enabled.
+  enabled: true
+
+  # Additional paths excluded from authentication.
+  # Health, docs, OpenAPI spec, and FDO protocol endpoints are always excluded.
```
knecasov: I would write "Health check" instead of "Health".

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,518 @@
...
+
+### Authenticator Interface
+
+Each auth mechanism implements this interface:
+
+```go
+type Authenticator interface {
+    // Name returns the mechanism name (e.g., "api-key", "oidc")
+    Name() string
+    // Authenticate inspects the request and returns an Identity if credentials are valid.
```
knecasov: Does it mean that Identity, nil is returned?

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2159 @@
...
+
+- Module path: `github.com/fido-device-onboard/go-fdo-server`
+- Existing GORM model pattern: models in `internal/state/`, init functions return `(*State, error)`, `AutoMigrate` called in init
+- Existing middleware pattern: `func(http.Handler) http.Handler` closures in `internal/middleware/`
+- Existing config pattern: structs with `mapstructure` tags in `internal/config/`, validated via `.Validate()` methods
+- Existing CLI pattern: Cobra commands in `cmd/`, Viper for config binding
+- OpenAPI specs: per-resource YAML in `api/v2/<resource>/openapi.yaml`, merged into `openapi.json` via `go:generate npx openapi-format`
+- FDO protocol endpoints (`/fdo/101/msg/*`, `/fdo/200/msg/*`), `/health`, `/api/docs/*`, `/api/openapi.json` are NEVER subject to management API auth
+- Default behavior when `auth` is not configured: all endpoints open (backward compatible)
+- SPDX header: `// SPDX-FileCopyrightText: (C) 2025 Red Hat Inc.` + `// SPDX-License-Identifier: Apache 2.0`
```
knecasov: Should not the year 2026 be used for new files, please?

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2159 @@
...
+### Task 1: Database Models for Users, Roles, and API Keys
+
+**Files:**
+- Create: `internal/state/user.go`
+- Create: `internal/state/role.go`
+- Create: `internal/state/apikey.go`
+- Create: `internal/state/auth.go`
+- Create: `internal/state/user_test.go`
+- Create: `internal/state/role_test.go`
+- Create: `internal/state/apikey_test.go`
```
knecasov, line 36: Why is not there a row "Create: `internal/state/auth_test.go`"?

knecasov (commented):
I added several comments.

Is there any reason for the `superpowers` subfolder, please? I would place `plans` and `specs` folders directly to `docs` folder.

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2159 @@
...
+
+	prefix := key[prefixStartIdx:prefixEndIdx]
+
+	candidates, err := state.FindAPIKeysByPrefix(a.db, prefix)
+	if err != nil {
+		return nil, auth.ErrInvalidCredentials
+	}
+
+	for _, candidate := range candidates {
+		if bcrypt.CompareHashAndPassword(candidate.HashedKey, []byte(key)) != nil {
```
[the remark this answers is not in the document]
mmartinv: Agreed — updated both the design spec and implementation plan to use SHA-256 with `crypto/subtle.ConstantTimeCompare` instead of bcrypt. API keys have ~190 bits of cryptographic entropy, so bcrypt's intentional slowness adds latency and DoS risk with no security benefit. The rationale is now documented in the spec's Storage and Validation section.

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2159 @@
...
+}
+
+func isExcluded(path string, excluded []string) bool {
+	for _, prefix := range excluded {
+		if strings.HasPrefix(path, prefix) {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
mmartinv: Fixed. The `isExcluded` function now uses exact match for leaf paths (e.g., `/health` matches only `/health`) and prefix match only for paths ending with `/` (e.g., `/api/docs/` matches `/api/docs/index.html`). This prevents bypass through path manipulation like `/health-check`. Also added path matching semantics documentation to the design spec.

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2159 @@
...
+		var roleNames []string
+		for _, r := range roles {
+			roleNames = append(roleNames, r.Name)
+		}
+
+		user, _ := state.GetUserByID(a.db, candidate.UserID)
+		name := ""
+		if user != nil {
+			name = user.Name
+		}
```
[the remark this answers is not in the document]
mmartinv: Fixed. The authenticator now verifies `User.Active` immediately after matching the key hash and before checking expiration. The design spec's authentication flow (step 6) now explicitly states: 'verify `User.Active = true` for the key's owning user — if user is inactive, return `nil, error`'.

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2159 @@
...
+	err := db.Where("prefix = ? AND active = ?", prefix, true).Find(&keys).Error
+	return keys, err
+}
+
+func UpdateAPIKeyLastUsed(db *gorm.DB, id string) {
+	now := time.Now()
+	if err := db.Model(&APIKey{}).Where("id = ?", id).Update("last_used_at", now).Error; err != nil {
+		slog.Warn("Failed to update API key last_used_at", "id", id, "error", err)
+	}
+}
```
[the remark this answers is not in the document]
mmartinv: Fixed. `UpdateAPIKeyLastUsed` now accepts `context.Context` and uses `db.WithContext(ctx)`. The caller passes `context.Background()` to the background goroutine (not the request context, which would be cancelled when the handler returns).

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2159 @@
...
+    authZ := auth.AuthZMiddleware(routeScopes)
+
+    mgmtHandlerV2 = middleware.RateLimitMiddleware(rate.NewLimiter(2, 10),
+        middleware.BodySizeMiddleware(10<<20,
+            authN(authZ(validationMiddleware(mgmtAPIServeMuxV2)))))
+} else {
+    mgmtHandlerV2 = middleware.RateLimitMiddleware(rate.NewLimiter(2, 10),
+        middleware.BodySizeMiddleware(10<<20,
+            validationMiddleware(mgmtAPIServeMuxV2)))
+}
```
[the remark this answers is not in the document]
mmartinv: Good catch — this was a critical gap. Added a 'V1 API Protection' section to the design spec requiring V1 endpoints to be wrapped with the same auth middleware when auth is enabled. Also added a note in the implementation plan's Task 8 reinforcing this requirement.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Fixed — updated to 3 operations (read, write, delete) to match the scope table.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Clarified. The authentication flow now explicitly checks both: step 4 queries candidates where `APIKey.Active = true`, and step 6 verifies `User.Active = true` for the owning user. A deactivated user's keys will be rejected.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Clarified — now reads: 'minimum 10 characters total — 4-char prefix + 6-char identifier minimum'.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Fixed — `rvinfo:write` description now reads 'Create/update rendezvous info config'.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Fixed — reformatted to: 'The semantic: `nil, nil` means "not my request, try the next authenticator"; `nil, error` means "credentials were present but invalid — stop and return 401".'

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Fixed — now reads 'Import/verify ownership vouchers' (no hyphen).

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Fixed — now reads '0 operations (public)'.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Fixed — now reads 'Health check, docs, OpenAPI spec, and FDO protocol endpoints are always excluded.'

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Clarified — the doc comment now explicitly states the return values: `(*Identity, nil)` on success, `(nil, nil)` to skip, `(nil, error)` on invalid credentials.

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
mmartinv: Fixed — all SPDX headers in the plan updated to 2026.

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
mmartinv: Fixed — added `internal/state/auth_test.go` to the file manifest in Task 1.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,526 @@
...
+### Authentication Flow
+
+1. Extract `X-API-Key` header — if absent, return `nil, nil` (try next authenticator)
+2. Validate format (`fdo_` prefix, minimum 10 characters total — 4-char prefix + 6-char identifier minimum) — if invalid, return `nil, error`
+3. Extract prefix (6 chars after `fdo_`)
+4. Query API key candidates by prefix where `APIKey.Active = true`
+5. Compute `sha256(provided_key)` and constant-time compare against each candidate's `HashedKey`
+6. If match found, verify `User.Active = true` for the key's owning user — if user is inactive, return `nil, error`
+7. Check expiration — if expired, log at debug level and return `nil, error` (same error as invalid credentials to avoid leaking key validity)
+8. Update `last_used_at` asynchronously (background goroutine with 5-second timeout via `context.WithTimeout`; tracked by `sync.WaitGroup` for graceful shutdown draining; failures are logged but do not block authentication)
```
djach7: Will we ever have a high enough auth load to worry about creating a goroutine per auth request and potential resource exhaustion? Or is that not really a concern?

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,526 @@
...
+```
+
+An API key can never exceed the user's permissions, even if the user's roles change after key creation.
+
+### Authentication Flow
+
+1. Extract `X-API-Key` header — if absent, return `nil, nil` (try next authenticator)
+2. Validate format (`fdo_` prefix, minimum 10 characters total — 4-char prefix + 6-char identifier minimum) — if invalid, return `nil, error`
+3. Extract prefix (6 chars after `fdo_`)
+4. Query API key candidates by prefix where `APIKey.Active = true`
```
djach7, line 235: Could this be combined with step 6 below? i.e. JOIN with User table and filter by User.Active in this query to avoid computing hashes for inactive users' keys

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
```diff
@@ -0,0 +1,526 @@
...
+5. Compute `sha256(provided_key)` and constant-time compare against each candidate's `HashedKey`
+6. If match found, verify `User.Active = true` for the key's owning user — if user is inactive, return `nil, error`
+7. Check expiration — if expired, log at debug level and return `nil, error` (same error as invalid credentials to avoid leaking key validity)
```
djach7: Do we ever cleanup the expired keys?

djach7 (commented):
Left some comments with a couple questions/ideas.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Not a practical concern. The goroutine is short-lived — a single DB UPDATE with a 5-second timeout, tracked by a `sync.WaitGroup` for graceful shutdown draining. Go handles millions of concurrent goroutines; this is one lightweight goroutine per authenticated request. The upstream rate limiter also caps request volume before auth is even reached.

If this ever became a concern (extremely high throughput), the mitigation would be batching `last_used_at` updates — collect key IDs in a channel and flush to DB in bulk every N seconds. But that's an optimization for a problem we don't have.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Valid optimization. A JOIN on the `users` table with `WHERE users.active = true` in the `FindAPIKeysByPrefix` query would skip hash computation for inactive users' keys entirely.

In practice the savings are negligible — with 6-char base62 prefixes (~56 billion combinations), the candidate set is almost always 1 row, so we'd save at most one SHA-256 computation (~microseconds). The current design keeps queries simple and separates concerns (key lookup vs user status check).

That said, if we later add key deactivation at scale or need to optimize the auth hot path, this JOIN is the right move. Noting it as a Phase 2 optimization candidate.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Not in Phase 1. Expired keys remain in the database but are rejected during authentication (the expiration check returns `ErrInvalidCredentials`), so they never grant access.

Over time they'll accumulate as dead rows. The cleanup mechanism belongs in Phase 2's management API — either:
- A `DELETE /api/v2/apikeys?expired=true` admin endpoint
- A periodic background cleanup job (e.g., delete keys expired > 30 days)

Adding a note to the design spec's future considerations to track this.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
djach7: Gotcha, that's kind of what I figured. Just wanted to double check, thank you!

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
djach7: Ah ok, sounds like a down the road change then if we get there. Thanks for the explanation.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
djach7: Noted, thank you

djach7 (dismissed):
LGTM

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
knecasov: Steps 2, 6, 7, and 10 all return errors but only step 7 specifies the logging level. It would be good to be explicit about logging for all of them for consistency and to make the implementer's job easier. What do you think about it, please?

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
knecasov: It is a detail, but I would switch the `Returns (nil, nil)...` and `Returns (nil, error)...` lines.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Good point. Added explicit logging levels to all error-returning steps in the authentication flow:

- Step 2 (invalid format): DEBUG
- Step 6 (inactive user): WARN
- Step 7 (expired key): DEBUG (already had this)
- Step 8 (last_used_at failure): WARN
- Step 10 (no match): DEBUG

The rationale: routine rejections (bad format, wrong key, expired) are DEBUG since they're expected in normal operation. Inactive user is WARN because it may indicate a revoked user still attempting access — worth operator attention.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Fixed — swapped the lines. The doc comment now reads error case first, then skip case:

```go
// Returns (nil, error) if credentials are present but invalid.
// Returns (nil, nil) if this authenticator doesn't match the request (no relevant credentials present).
```

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
knecasov: Log levels make sense. Minor note for documentation purposes: "Step 8 (last_used_at failure): WARN" is correct because it is a server-side failure (DB write) on an otherwise successful authentication path — not a client error. Worth operator attention, while not blocking the request.

Do you plan to incorporate the logging levels into the implementation plan?

knecasov (commented):
I added my comments. I would also like to remind this [general comment](https://github.com/fido-device-onboard/go-fdo-server/pull/243#pullrequestreview-4615897546) from my side.

mmartinv:
> I added several comments.
> 
> Is there any reason for the `superpowers` subfolder, please? I would place `plans` and `specs` folders directly to `docs` folder.

I used the `superpowers` Claude plugin to generate both the specification and the implementation plan. This is the plugin's default directory, and I think it's a good choice since it's under docs. Keeping these documents in their own dedicated directory also makes it easier to identify which documents were AI-generated.

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
mmartinv: Yes — just pushed a commit that incorporates the logging levels into the implementation plan. Task 3 (API Key Authenticator) now has:

1. A **logging levels reference table** in the task description summarizing the level and rationale for each error-returning step.
2. **Inline `slog` calls** in the `Authenticate` method code matching the design spec:
   - `slog.Debug` for invalid format, expired key, and no-match (routine rejections)
   - `slog.Warn` for inactive user (potential revoked-user access attempt)
   - `slog.Warn` for `last_used_at` update failure (already present — server-side DB failure on successful auth path)

This way the implementer has the levels right alongside the code without needing to cross-reference the design spec.

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2335 @@
...
+func AuthZMiddleware(routeScopes map[string][]string) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			identity := IdentityFromContext(r.Context())
+			if identity == nil {
+				writeAuthError(w, http.StatusUnauthorized, "authentication required")
+				return
+			}
+
+			required := lookupRequiredScopes(r.Method, r.URL.Path, routeScopes)
```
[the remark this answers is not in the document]
mmartinv: Fixed. The AuthZ middleware now cleans the request path with `path.Clean(r.URL.Path)` before passing it to `lookupRequiredScopes`. This prevents authorization bypass via double slashes or dot segments (e.g., `/vouchers//123` no longer evades pattern matching).

Updated both the design spec (AuthZ middleware description now includes path cleaning as explicit step 1) and the implementation plan (Task 4 middleware code).

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2335 @@
...
+
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+func isExcluded(path string, excluded []string) bool {
+	for _, entry := range excluded {
+		if strings.HasSuffix(entry, "/") {
+			if strings.HasPrefix(path, entry) {
```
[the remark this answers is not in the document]
mmartinv: Fixed. The `isExcluded` prefix matching now uses `strings.HasPrefix(cleanedPath+"/", entry)` instead of `strings.HasPrefix(cleanedPath, entry)`. This ensures that `path.Clean`'s trailing-slash stripping does not break prefix matching — e.g., a request to `/api/docs/` is cleaned to `/api/docs`, then `"/api/docs/" HasPrefix "/api/docs/"` correctly matches.

Updated the design spec's path matching semantics section and the implementation plan (Task 4 `isExcluded` function).

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2335 @@
...
+func (a *APIKey) ScopesList() []string {
+	if a.Scopes == "" {
+		return nil
+	}
+	var scopes []string
+	if err := json.Unmarshal([]byte(a.Scopes), &scopes); err != nil {
+		return nil
+	}
+	return scopes
+}
```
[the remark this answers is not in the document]
mmartinv: Good catch — this is a privilege escalation via fail-open. Fixed.

`ScopesList()` now returns `([]string, error)`. If `json.Unmarshal` fails on corrupted data, the error propagates up through `resolveScopes` to the authenticator, which returns a non-`ErrInvalidCredentials` error. The AuthN middleware then returns 500 (not 401), and no scopes are granted.

Added a "Fail-closed on corrupted scopes" section to the design spec's Scope Resolution Logic, and updated the implementation plan in Task 1 (`ScopesList` signature + test) and Task 3 (`resolveScopes` error handling).

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2335 @@
...
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			for _, authn := range authenticators {
+				identity, err := authn.Authenticate(r.Context(), r)
+				if err != nil {
+					slog.Debug("Authentication failed", "authenticator", authn.Name(), "error", err)
+					writeAuthError(w, http.StatusUnauthorized, "authentication failed")
+					return
```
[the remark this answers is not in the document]
mmartinv: Fixed. The AuthN middleware now classifies errors:

- `errors.Is(err, ErrInvalidCredentials)` → 401 Unauthorized (bad credentials, logged at DEBUG)
- Any other error (DB failures, scope corruption, etc.) → 500 Internal Server Error (logged at ERROR)

This also required updating the `APIKeyAuthenticator` to return distinct error types: `auth.ErrInvalidCredentials` for credential problems (invalid format, inactive user, expired key, no match) and `fmt.Errorf(...)` for infrastructure problems (DB lookup failures, scope parsing errors). The middleware uses `errors.Is` to distinguish them.

Updated both the design spec (AuthN middleware step 4 now documents the error classification) and the implementation plan (Task 3 authenticator + Task 4 middleware + new test `TestAuthNMiddleware_InternalError`).

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2335 @@
...
+			slog.Debug("Registered route scope", "route", key, "scopes", scopeSlice)
+		}
+	}
+
+	return routeScopes, nil
+}
+
+func parseStringSliceExtension(ext interface{}) ([]string, error) {
+	raw, ok := ext.([]interface{})
+	if !ok {
```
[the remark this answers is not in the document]
mmartinv: Fixed. `parseStringSliceExtension` now checks for `[]string` first before falling through to `[]interface{}`. This handles the case where the extension value is pre-parsed (e.g., in unit tests) or comes in as a native string slice.

Updated the implementation plan (Task 5 `scopes.go`).

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2335 @@
...
+    }
+    apiKey, err := state.SeedAuth(ctx, gormDB, *config.Auth.Seed, mfgScopes, false)
+    if err != nil {
+        return nil, fmt.Errorf("failed to seed auth: %w", err)
+    }
+    if apiKey != "" {
+        fmt.Fprintf(os.Stderr, "WARNING: Initial admin API key generated — save this key, it will not be shown again:\n%s\n", apiKey)
+    }
+}
+```
```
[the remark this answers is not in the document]
mmartinv: Valid design gap — fixed.

The authenticator is now created at the server startup level (in `NewManufacturingServer` etc.), not inside `Handler()`. The server retains a reference to the `*APIKeyAuthenticator` and calls `Wait()` during graceful shutdown after `httpServer.Shutdown(ctx)` returns. `Handler()` receives the pre-built authenticator slice as a parameter.

Updated both the design spec (step 8 of the authentication flow now documents WaitGroup exposure and the requirement that the server retains the authenticator reference) and the implementation plan (Task 8 Step 1 — `Handler()` accepts `authenticators []auth.Authenticator`; Step 2 — authenticator created at server level with shutdown draining).

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2390 @@
...
+```go
+var mgmtHandlerV1 http.Handler
+if authCfg.Enabled {
+    mgmtHandlerV1 = authN(mgmtAPIServeMuxV1)
+} else {
+    mgmtHandlerV1 = mgmtAPIServeMuxV1
+}
+```
+
+Note: V1 endpoints do not need `authZ` (scope checking) since they predate the scope model — `authN` alone ensures only authenticated users can access them. Scope-based authorization for V1 can be added if V1 is retained long-term.
```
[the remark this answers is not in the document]
mmartinv: Fixed — V1 endpoints are now wrapped with both `authN(authZ(...))`, not just `authN`.

This works with zero additional scope mapping because V1 routes are served behind `http.StripPrefix("/api/v1", ...)`. By the time the AuthZ middleware sees the request, the path matches the same patterns as V2 (e.g., `GET /vouchers`, `POST /rvinfo`), so the same `routeScopes` map covers both versions.

Updated both the design spec (V1 API Protection section) and the implementation plan (Task 8 Step 1).

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2390 @@
...
+
+- [ ] **Step 2: Update server startup to init auth DB and seed**
+
+In `internal/server/manufacturing.go`, update `NewManufacturingServer` after DB init:
+
+```go
+if err := state.InitAuthDB(ctx, gormDB); err != nil {
+    return nil, fmt.Errorf("failed to initialize auth database: %w", err)
+}
+
```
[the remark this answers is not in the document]
mmartinv: Fixed. `InitAuthDB` is now wrapped in `if config.Auth.Enabled` — no DDL operations when auth is disabled.

Updated the design spec (Server Startup Sequence, Migration Path, Database Migration sections) and the implementation plan (Task 8 Step 2).

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2390 @@
...
+		seedCfg := config.SeedConfig{
+			Admin: &config.SeedAdminConfig{Name: name, Email: email},
+		}
+		apiKey, err := state.SeedAuth(ctx, gormDB, seedCfg, serverScopes, force)
+		if err != nil {
+			return err
+		}
+
+		fmt.Println(apiKey)
+		return nil
```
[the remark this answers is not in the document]
mmartinv: Fixed. The `init-admin` command now returns a non-zero exit code with a clear error message when users already exist:

```go
if apiKey == "" {
    return fmt.Errorf("auth database already seeded; use --force to create an additional admin user")
}
```

Updated the design spec (CLI Bootstrap Command section) and the implementation plan (Task 9 Step 1).

docs/superpowers/plans/2026-06-29-management-api-auth-phase1.md
```diff
@@ -0,0 +1,2390 @@
...
+	}
+
+	if !a.HasEnabledMechanism() {
+		return errors.New("auth is enabled but no authentication mechanism is configured")
+	}
+
+	slog.Debug("Auth configuration validated", "mechanisms", a.enabledMechanismNames())
+	return nil
+}
```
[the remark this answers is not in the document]
mmartinv: Fixed. `Validate()` now checks that all `ExcludedPaths` entries start with `/`:

```go
for _, p := range a.ExcludedPaths {
    if p == "" || !strings.HasPrefix(p, "/") {
        return fmt.Errorf("invalid excluded path %q: must start with '/'", p)
    }
}
```

Updated the design spec (Validation Rules section) and the implementation plan (Task 6 — `AuthConfig.Validate()` + two new tests `TestAuthConfig_Validate_InvalidExcludedPath` and `TestAuthConfig_Validate_ValidExcludedPath`).

docs/superpowers/specs/2026-06-29-management-api-auth-phase1-design.md
knecasov: Thank you very much for updating the implementation plan!

djach7 (approved):
Re-reviewed after force push. 

Design spec (632 lines): Read in full. All feedback from previous review rounds has been incorporated, including the auth flow order correction (expiry before user.Active) and the nil identity guard on ContextWithIdentity.

Implementation plan (3554 lines, 10 tasks): Read end-to-end. Cross-referenced Tasks 1-3 against the actual implementation in #244 — the code matches the plan character-for-character. The remaining tasks (4-10) are internally consistent with the spec and with each other. Key things verified:

- Error contract (ErrInvalidCredentials → 401, internal → 500) is consistent across spec, plan, and implementation
- V1 API protection is explicitly addressed with per-server scope overrides (manufacturer: 1, owner: 8, rendezvous: 0)
- path.Clean used in both AuthN and AuthZ for bypass prevention
- Seed logic is atomic (single transaction) with preset API key support
- Graceful shutdown drains apikeyAuthn.Wait() after httpServer.Shutdown
- SPDX headers use 2026 throughout

knecasov (approved):
I noticed one detail - I would unify the order of the return values for the `Authenticate` function. I would mention `(nil, error)` first, followed by `(nil, nil)`.

Anyway, I do not request this change and LGTM otherwise (changes from PR #244 should be incorporated in this PR), so I am approving it.
Thank you very much for incorporating my notes!

PR #248, opened 2026-07-29
feat: add auth middleware and OpenAPI scope parser (THEEDGE-4820)

mmartinv:
## Summary

Implements the AuthN/AuthZ middleware chain and OpenAPI scope parser for management API authentication (THEEDGE-4820 / Plan Tasks 4-5):

- **AuthN middleware** — iterates authenticators in order, classifies errors (`ErrInvalidCredentials` → 401, other → 500), supports path exclusion for FDO protocol, health, and docs endpoints with `path.Clean` normalization
- **AuthZ middleware** — checks `Identity` scopes against per-route requirements using segment-by-segment path matching with `{param}` wildcard support
- **OpenAPI scope parser** — `ParseRouteScopes` extracts `x-required-scopes` extensions from the merged OpenAPI JSON spec at startup, normalizing `/api/v2/` prefix; `CollectAllScopes` aggregates unique scopes for seed/bootstrapping

## Key design decisions

- **Path exclusion semantics** — exact match for leaf paths (`/health`), prefix match for `/`-terminated paths (`/api/docs/`); `path.Clean` applied before matching to prevent bypass via `//` or `..` segments
- **Error classification** — `errors.Is(err, ErrInvalidCredentials)` traverses the error chain, so authenticators can wrap the sentinel; non-credential errors return 500 to prevent masking DB failures as client errors
- **JSON error responses** — `authErrorResponse` matches the existing `errorResponse` pattern in `internal/middleware/validation.go`
- **Scope parser path normalization** — `strings.CutPrefix(path, "/api/v2/")` with trailing slash to prevent false matches on paths like `/api/v2extra`
- **Route matching** — O(n) scan over route scope entries with segment-count guard; `{param}` segments match any value; acceptable given ~17 operations across all specs

### New
- `internal/auth/middleware.go` — AuthNMiddleware, AuthZMiddleware, path exclusion, route matching, JSON error writer
- `internal/auth/middleware_test.go` — 17 tests (excluded paths, auth chain, error classification, response format, context propagation, path traversal)
- `internal/auth/scopes.go` — ParseRouteScopes, CollectAllScopes, x-required-scopes extension parser
- `internal/auth/scopes_test.go` — 4 tests (scope extraction, no-extension, multiple scopes, collect-all)

## Test plan
- [x] All 21 new tests pass
- [x] Full test suite passes (no regressions across 9 packages)
- [x] `go vet ./...` clean
- [x] Each commit builds and tests independently

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/auth/middleware.go
```diff
@@ -0,0 +1,156 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			identity, _ := IdentityFromContext(r.Context())
+			if identity == nil {
+				writeAuthError(w, http.StatusUnauthorized, "authentication required")
+				return
+			}
+
+			required := lookupRequiredScopes(r.Method, path.Clean(r.URL.Path), routeScopes)
+			if len(required) == 0 {
```
say-paul, line 58: This will allow silent pass through if new endpoint is added without : `x-required-scopes`,  endpoint can be accessed without requiring auth.

mmartinv: That's the expected behavior, the source of truth is the Open API spec and to protect a URL you need to add `x-required-scopes` to each API endpoint operation.

However, your question makes me wonder if there's a better way of wiring the default exceptions like the fdo endpoints. Also, I believe the '/health' shouldn't be there as it's already in the Open API spec.

internal/auth/middleware.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func lookupRequiredScopes(method, path string, routeScopes map[string][]string) []string {
+	for pattern, scopes := range routeScopes {
+		if matchRoute(method, path, pattern) {
+			return scopes
+		}
+	}
+	return nil
+}
```
knecasov: The map iteration order in `lookupRequiredScopes` is non-deterministic, which creates an authorization bypass when multiple route patterns match the same request (for example, theoretically,`GET /vouchers/{guid}` and `GET /vouchers/summary` both match the path `/vouchers/summary`). Consider implementing deterministic precedence by collecting all matches and returning the one with the fewest wildcard segments. This prevents intermittent, untestable authorization bypass.

knecasov (commented):
I added one comment that does not block merging.

Failed checks seem to be unrelated to the changes. Could you please mention it in the PR if you agree?

internal/auth/middleware.go
mmartinv: I've addressed both points raised here:

- Removed the hardcoded `defaultExcludedPaths` from the auth package. The caller now explicitly declares which paths bypass AuthN when wiring the middleware, making the public surface visible at the integration site.
- Added `ParsePublicPaths()` which extracts paths from the OpenAPI spec that have `security: []` (the standard OpenAPI way to mark an endpoint as public). This means `/health` no longer needs to be hardcoded — it's derived from the spec itself.

Non-spec paths like the FDO protocol endpoints are still passed manually by the caller.

internal/auth/middleware.go
mmartinv: Good catch — fixed. `lookupRequiredScopes` now collects all matching patterns and returns the most specific one (fewest wildcard segments), with alphabetical tie-breaking for full determinism. Added a test (`TestAuthZMiddleware_LiteralRoutePreferredOverParameterized`) that registers both `GET /vouchers/{guid}` and `GET /vouchers/summary` and verifies the literal route wins.

mmartinv:
Thanks for the review! I've addressed the inline comment about non-deterministic matching. Regarding the failing checks — they are unrelated to this PR's changes (pre-existing CI flakiness).

knecasov (approved):
LGTM, thank you very much for changes!

runcom (approved):

